Issue #11, opened 2021-11-22
Loading model libraries doesn't work on platforms other than Linux


hannobraun:
Right now, the host application expects the dynamic library to end with `.so`, which is correct on Linux. It should either fall back to other file endings to support more platforms, or there might be a way to override the file name outright, and set it to something like `.model`.

hannobraun:
In case anyone wants to take a look, the library is loaded in [`model.rs`](https://github.com/hannobraun/fornjot/blob/06d90d846c38249466409a73e3a398f3d46de49f/src/model.rs#L43). If it's practical to override the whole file name, this [also needs to happen in `model.rs`](https://github.com/hannobraun/fornjot/blob/06d90d846c38249466409a73e3a398f3d46de49f/src/model.rs#L30-L33)

hannobraun:
I initially thought of this as an enhancement (as in, it works fine, it should work fine in more places), but I think it makes more sense to classify it as a bug.

Bandsberg:
For OSX: dynamic library: .dylib (e.g. libspacer.dylib) results in "Error: Error request adapter". Error defined in renderer.rs
Will continue to investigate next time I have time

hannobraun:
@Bandsberg I think what you're seeing is most likely #10, not this issue.

hannobraun:
Fixed by @Bandsberg in #23.

PR #23, opened 2021-12-17
Os based lib import path string

Bandsberg:
Simple fix for #11. Seems a bit hacky. Might not be super robust?

Tested on macOS and Windows 10. @hannobraun will you test on Linux?

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -39,11 +39,27 @@ impl Model {
         // TASK: Read up why those calls are unsafe. Make sure calling them is
         //       sound, and document why that is.
         let shape = unsafe {
-            let lib = libloading::Library::new(format!(
-                "{}/target/debug/lib{}.so",
-                self.path(),
-                self.name(),
-            ))?;
+            let lib = libloading::Library::new(
+                if cfg!(windows) {
+                    format!(
+                        "{}/target/debug/{}.dll",
+                        self.path(),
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thanks again, @Bandsberg! Making these fixes and testing on platforms I don't have access to is very appreciated!

I just did a quick test, and it works perfectly on Linux. It would be a bit nicer, if the `if` expression just returned the file ending, so the `format!` wouldn't be duplicated for each platform, but this will do for now.

(I do have a rule to keep me sane: It's always okay to write messy code, but it's forbidden to build on top of code that's already messy. This rule has been working exceptionally well for me. Keeps things moving while making sure the mess always stays manageable.)

hannobraun:
> Seems a bit hacky. Might not be super robust?

Oh, I meant to comment on that part, but forgot. I think it's fine. It's an improvement over what was there before, and that's what counts. We can always do better, of course, but that shouldn't prevent us from keeping things moving.

In any case, it works for all the platforms that we seem to have access to. If someone wants to compile this on a BSD or something else, they're welcome to open issues and submit PRs. That why it's open source :smile:

Bandsberg:
Happy to help. Feedback on PRs always much appreciated :)
I might spend a little time making it nicer. I fully agree with your thoughts on continues improvement.

PR #26, opened 2021-12-21
Cleanup: Cleaner code of lib path

Bandsberg:
Nicer code for the path to imported lib, As discussed in PR #23.
Untested for Linux.

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -39,13 +39,18 @@ impl Model {
         // TASK: Read up why those calls are unsafe. Make sure calling them is
         //       sound, and document why that is.
         let shape = unsafe {
-            let lib = libloading::Library::new(if cfg!(windows) {
-                format!("{}/target/debug/{}.dll", self.path(), self.name(),)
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @Bandsberg!

I tested on Linux, works great! I also rebased this branch on top of `main`, to clean up the history, and pushed the commit manually (https://github.com/hannobraun/fornjot/commit/6ead39a7fe3de60e14aa1539b239d8a598aeb017).

Some notes for potential further clean-ups:
- It's a bit weird to have this whole code block as an argument to `Library::new`. I think it makes sense to assign the result to to a variable, then just pass that to `new`.
- As a complement to the previous point, none of that formatting code is unsafe anyway, so it makes sense to move it in front of the `unsafe` block.
- I think it would even be a bit nicer, to merge that first and last `format!`. No reason to format `path` separately, I think.

Just some notes, in case you or anyone else want to put more work into this.

PR #27, opened 2021-12-26
Lib path cleanup - Moving code to method

Bandsberg:
This is for me the most readable/expressive way of doing this. Let me know what you think 😄


hannobraun (approved):
Thank you, looks great!

Issue #12, opened 2021-11-22
Reloading the model doesn't work on macOS


hannobraun:
I've gotten a report that reloading the model doesn't work on MacOS. I'm just using [Notify](https://crates.io/crates/notify), so maybe it's an issue in that library.

I don't have a Mac, and no recent experience with MacOS, so any help here would be appreciated. As of this writing, the code for that lives in [`main.rs`, line 56 and following](https://github.com/hannobraun/fornjot/blob/06d90d846c38249466409a73e3a398f3d46de49f/src/main.rs#L56), but it'll probably move somewhere else at some point (likely either `model.rs` or a dedicated module).

Bandsberg:
Having tested this on Windows10, the issue is present there aswell.
According to documentation of Notify's AccessKind:

>  Only some platforms are capable of generating these.

So it makes sense that it does not work for Window10 and MacOS. The documentation does not state which platforms can, and cannot use AccessKind.

A different point relating to a potential fix. I suppose the desired behavior is to recompile on a save event. As I understand Notify documentation currently the recompile is on:

> An event describing non-mutating access operations on files

I will look in to a fix using a 'saving of file' event. @hannobraun let me know if I am going in the wrong direction.

FYI. I used debugging CodeLLDB VScode plugin to debug and learned that debugging and the recompiling functionality does not work well together. I believe it is due to the debugger restraining access to the compiled libraries (e.g. xzy.dll) so that they cannot be overwritten. Not an issue for this fix, but an interesting issue nonetheless.

hannobraun:
Thanks for looking into this, @Bandsberg. It sounds to me like you're going in the right direction.

I implemented this (on Linux) without looking at the documentation at all. I just printed the events that were triggered and selected the one that seemed to do what I needed, naively assuming that `notify` behavior was uniform across platforms.

Maybe another event works cross-platform. If not, we can add platform-specific code for this.

> FYI. I used debugging CodeLLDB VScode plugin to debug and learned that debugging and the recompiling functionality does not work well together. I believe it is due to the debugger restraining access to the compiled libraries (e.g. xzy.dll) so that they cannot be overwritten. Not an issue for this fix, but an interesting issue nonetheless.

Feel free to open an issue, if you think there's something we can do about this on our end. It should be relatively straight-forward to make the automatic re-compile something that can be enabled/disabled. And once this is possible, we could let the user do this through the UI, or disable it as part of error handling. I don't know, whatever makes sense.

[...]

PR #28, opened 2021-12-27
Model recompile on windows and MacOS

Bandsberg:
Changes recompile to 'save file event'. 

Solves some of the issues related to #12, as the model recompiles and reloades on Win10 but only recompiles on MacOS.
The issue of reloading model on MacOS does not appear to be related monitoring of/watching for changes to model files.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,45 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "lldb",
+            "request": "launch",
+            "name": "Debug executable 'fj-host'",
+            "cargo": {
+                "args": [
+                    "build",
+                    "--bin=fj-host",
+                    "--package=fj-host"
+                ],
+                "filter": {
+                    "name": "fj-host",
+                    "kind": "bin"
+                }
+            },
+            "args": [],
+            "cwd": "${workspaceFolder}"
+        },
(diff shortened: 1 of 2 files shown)


Bandsberg:
Also added Debug settings for codeLLDB VSCode plugin. Maybe that should either be removed or added to VScode as a suggested plugin?

hannobraun (approved):
Thank you, @Bandsberg, looks great! I also just tested this on my machine, and can confirm that it still works on Linux.

There's one problem: Two rebuilds are triggered for the same modification. The second rebuild barely takes any time though, so it's only a minor annoyance. And that it works on more platforms is more important anyway.

I'm going to merge this and will open a separate issue for the multiple-rebuilds problem.

hannobraun:
> Also added Debug settings for codeLLDB VSCode extension. Maybe that should either be removed or added to VScode as a suggested extension?

Okay, as usual I forgot to reply to something.

I'm not using a debugger right now, but I'm fine with merging configuration that is used by regular contributors, as long as it's not too disruptive otherwise.

By "suggested extension", do you mean [Recommended Extensions](https://code.visualstudio.com/docs/editor/extension-marketplace#_recommended-extensions)? I haven't used this feature before, so I'm not sure what to do here. Would that mean that every VS code user would get a pop-up about this extension, or something? If so, I don't think this is necessary.

Bandsberg:
> By "suggested extension", do you mean [Recommended Extensions](https://code.visualstudio.com/docs/editor/extension-marketplace#_recommended-extensions)? I haven't used this feature before, so I'm not sure what to do here. Would that mean that every VS code user would get a pop-up about this extension, or something? If so, I don't think this is necessary.

I do and yes there would be a pop-up. Roger that 😄

Issue #29, opened 2021-12-28
Changing the model file triggers multiple rebuilds


hannobraun:
Changing the model file while the host is running can trigger multiple rebuilds. I'm seeing two on Linux, and don't know about other platforms.

This is only a minor annoyance, so it's not critical to fix this. It would be nicer though if every change triggered exactly one rebuild, to save unnecessary work. This might be possibly by choosing the events we listen for more carefully, or maybe some kind of de-duplicating feature is required.

hannobraun:
Adding the `good first issue` label, since this is a fairly isolated problem that doesn't affect a lot of code. As of this writing, [all the relevant code](https://github.com/hannobraun/fornjot/blob/591a5acf2017d06002dc12336fdb6509832b5070/src/main.rs#L59-L99) is located in `main.rs`. I think it would make sense to move it to `model.rs` though, but this doesn't need to be a part of this fix.

mxdamien:
The watcher is watching the complete 'mode/{model_name}/src' subfolder for changes. Events will be triggered for lib.rs but also for temporary files (e.g. Vim swap files during editing without saving). Would it make sense to restrict the watcher to just changes to model/{model_name}/src/lib.rs ?

hannobraun:
Thank you, for the suggestions, @mxdamien!

I'm not sure what solution would be best. The intention is that a models can written in Rust, not just some restricted dialect. So watching just `lib.rs` would be too restrictive, as a non-trivial model could easily consist of multiple modules. Watching for all `*.rs` files would be better, but someone could include some non-Rust data file in their model (through something like [`include_str!`](https://doc.rust-lang.org/std/macro.include_str.html)/[`include_bytes!`](https://doc.rust-lang.org/std/macro.include_bytes.html)).

Maybe I'm being too paranoid about this last one, but I've experienced time and time again that software that's trying to be clever without being clever enough will become a huge frustration at some point.

How about this idea:
- Watch all the Cargo files that could influence a build (`Cargo.toml`, `Cargo.lock`, `build.rs`, `.cargo/config`, `.cargo/config.toml` are the ones I can think of off the top of my head).
- In addition, watch _everything_ under `src/`.
- Filter out events according to some built-in blacklist. That would include things like Vim swap files.

Would that work, or am I missing something? Maybe it would be better to leave out that first point, just watch everything and have a blacklist.

mxdamien:
[...]

PR #39, opened 2022-01-19
Match file extensions against blacklist to avoid multiple rebuilds tr…

mxdamien:
…iggered by temporary files

see #29

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,6 +9,8 @@ mod mesh;
 mod model;
 mod window;
 
+use std::collections::HashSet;
+use std::ffi::OsStr;
 use std::{collections::HashMap, sync::mpsc, time::Instant};
 
 use futures::executor::block_on;
@@ -139,6 +141,24 @@ fn main() -> anyhow::Result<()> {
                 notify::event::DataChange::Content,
             )) = event.kind
             {
+                let file_ext = event
+                    .paths
+                    .get(0)
+                    .expect("File path missing in watch event")
+                    .extension();
+
+                let black_list = HashSet::from([
+                    OsStr::new("swp"),
+                    OsStr::new("tmp"),
+                    OsStr::new("swx"),
+                ]);
+
+                if let Some(ext) = file_ext {
+                    if black_list.contains(ext) {
+                        return;
+                    }
+                }
+
                 let shape = match model.load(&parameters) {
                     Ok(shape) => shape,
                     Err(model::Error::Compile) => {


src/main.rs
```diff
@@ -139,27 +141,37 @@ fn main() -> anyhow::Result<()> {
...
-                        panic!("Error reloading model: {:?}", err);
-                    }
-                };
-
-                // This will panic, if the other end is disconnected, which is
-                // probably the result of a panic on that thread, or the
-                // application is being shut down.
-                //
-                // Either way, not much we can do about it here, except maybe to
-                // provide a better error message in the future.
```
hannobraun, line 161: You're removing this comment, even though it still applies in the changed version. Can you please re-add it?

src/main.rs
```diff
@@ -139,27 +141,37 @@ fn main() -> anyhow::Result<()> {
...
+
+                let black_list = HashSet::from([
+                    OsStr::new("swp"),
+                    OsStr::new("tmp"),
+                    OsStr::new("swx"),
+                ]);
+
+                if (file_ext.is_some()
+                    && black_list.get(file_ext.unwrap()).is_none())
+                    | file_ext.is_none()
```
hannobraun, line 158: This is just a nit-pick. Feel free to act on it (or comment, if you disagree), or just leave it as-is, as you prefer. It doesn't matter a whole lot.

That `file_ext.is_some()` followed by `file_ext.unwrap()` is a bit icky. `if let Some(ext) = file_ext` would be more robust, but can't be used as a direct replacement here.

Alternative approach:
``` rust
if let Some(ext) = file_ext {
    if black_list.contains(ext) {
        return;
    }
}

// insert rest of event handling here
```

Aside from using the more robust `if let`, this also makes the handling of the error case a self-contained thing, thereby making the code a bit easier to understand. It also saves one level of indentation for the happy path. But as I said, not that important, feel free to ignore.

hannobraun (changes requested):
Thank you for the pull request, @mxdamien!

I found one problem (code comment went missing; see the review comment I left) and left another note to highlight a possible improvement.

src/main.rs
mxdamien: Thanks for the feedback.

Removing the comment was an accident - sorry for that.

The alternative approach indeed looks neater. Still getting used to the rust coding style coming from C/C++.
I've pushed the changes

hannobraun: Hey, no problem, that what reviews are for! :smile:

hannobraun (approved):
Thank you for the changes, @mxdamien. Looks great!

PR #50, opened 2022-01-20
Add star model

hannobraun:
I wanted to add an example model with a star shape, both to test/demonstrate support for concave sketches and because I think the code that generates the star shape serves as a nice demonstration for the kind of thing that Code-CAD is useful for.

star

Results are decidedly mixed :smile: 

I haven't had a chance to look into it yet in any detail, although I'm pretty sure this displays two distinct problems:
- Not all triangles are generated counter-clockwise.
- Some triangles are discarded erroneously.

I'll look into it when I get a chance and will open/fix issues as required.


hannobraun:
I've opened #59 and #60. Already fixed #59 in #61. Here's what the model looks like now:

Screenshot from 2022-01-20 20-08-33

Slightly better. Still working on #60.

hannobraun:
Looks correct now, after #81.

Screenshot from 2022-01-25 18-05-02

This is not merge-ready though. I still want to put a star-shaped hole into it.

hannobraun:
And done!

Screenshot from 2022-01-26 14-49-14

Issue #78, opened 2022-01-24
Store topological relationships, cache geometrical data


hannobraun:
I'm not completely sure I'm using these mathematical terms correctly, so let me provide some examples.

Examples of topological relationships:
- The vertices that bound this edge
- The edges which bound this face
- The neighboring faces of the edge

Examples of geometrical data:
- The point at (0, 0, 0)
- The line segment from (1, 2, 3) to (4, 5, 6)

Right now, we're storing multiple copies of geometrical data in various locations. For example, a vertex that is shared between two edges is stored, by value, in both of these edges. More than likely, it's also been computed multiple times before it was stored. Thanks to the well-known issues around floating point accuracy, this means it might actually have been computed differently each time.

This can lead to bugs, like the one fixed in #74. This pull request demonstrates how these kinds of issues can be addressed: By being careful when comparing floating point numbers, to make sure slightly different values that are supposed to be the same, are recognized as the same.

That approach is not great. For such comparisons, a suitable epsilon value must be chosen. This is bound to lead to problems somewhere down the line, when a use case that was not foreseen produces values that the chosen epsilon is not appropriate for. That could lead to values that are supposed to be different to instead be recognized as the same, which sounds like fuel for nightmare bugs. It's also hard to find all these cases. If we were to do it like that, we could expect a long period of plugging holes as users (hopefully) report them.

I think a better approach would be to prevent this kind of thing from happening at an architectural level. Here's what I have in mind:

- There is only one place where geometrical data is stored: A centralized cache.
- The values in that cache are referred to using ids.
- Those ids are what is actually stored in other data structures.
- If multiple locations refer to the same vertex, they do so using the same id, not multiple values that might or might not be slightly different.

To give an example:

1. A model defines a sketch as a series of points.
2. The code that processes this sketch stores these points in the vertex cache, getting ids back.
3. That same code, when it creates the edges that make up the sketch, gives those ids to the edges. The edges themselves never know the actual locations of those points, and don't need to.
4. If another piece of code needs those values (like, for example, to triangulate a face), it is provided with a reference to the cache and can look those values.

For now, all of this would happen in the host application. At a later point, it makes sense to move some of this into the `fj` library so models can refer to geometry they previously generated. But that can happen later.

This is going to be a large-scale refactoring effort, and I'm sure I haven't completely thought through all the details. I still wanted to open this issue, to provide some context on the existing code, and the work that's about to happen.

hannobraun:
[...]

PR #110, opened 2022-01-28
Add geometry cache

hannobraun:
This is an early in-progress snapshot of my work on #78. Not sure if this is the right approach. I'm mostly sketching at this point, and there quite a few questions that still need to be answered.

I plan to keep pushing my work-in-progress here.


hannobraun:
Another, much simpler, approach was chosen, so this is no longer necessary. See #78 for context. Closing.

PR #111, opened 2022-01-28
Fix #91 

### LINKED ISSUE #91: Circle approximation can freeze host application for extended time
The circle approximation code contains [a loop](https://github.com/hannobraun/Fornjot/blob/8732095651025746088c16deff85eb37b3e2d83c/src/kernel/geometry/curves/circle.rs#L41-L61) that can become long-running, if the `tolerance` parameter is unreasonably low. This happened once before, due to a bug in calculating `tolerance`.

I think that we can do away with that loop completely by being a bit more clever about calculating `n`, which would solve the problem. I haven't actually sat down and done the math though, so I'm not 100% sure.

If it turns out this is not possible, then an alternative solution would be to create some system that can monitor and abort long-running algorithms. We might need something like that anyway, long-term, but that one loop is not a good enough reason to work on that now.

Labeling as `good first issue`, as this is basically a math problem, and the affected code is fully contained within a reasonable method.

gzsombor:
with an exact formula to calculate the number of vertices needed to approximate a circle

### DIFF
--- a/src/kernel/geometry/curves/circle.rs
+++ b/src/kernel/geometry/curves/circle.rs
@@ -34,31 +34,7 @@ impl Circle {
         // and the circle. This is the same as the difference between
         // the circumscribed circle and the incircle.
         //
-        // Let's figure which regular polygon we need to use, by just
-        // trying out some of them until we find one whose maximum error
-        // is less than or equal to the tolerance.
-        let mut n = 3;
-        loop {
-            let incircle_radius = radius * (PI / n as f64).cos();
-            let maximum_error = radius - incircle_radius;
-
-            if maximum_error <= tolerance {
-                break;
-            }
-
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @gzsombor, this is awesome! I'm really enjoying getting rid of that crummy loop :smile: 

This looks great. I'm going to merge it. Especially good job on the testing!

I've left a bunch of comments, but all of those are just nitpicking on a high level. Definitely none of them are a blocker, and I'm going to merge the pull request right away. If you want to, feel free to address those in a follow-up PR. (Or if anyone else happens to come by, wanting to help out, that goes for you too.)

gzsombor:
Thanks for the comments, yes, you're right, the self is not needed, and u32 is better type for counting things.

PR #114, opened 2022-01-31
Use caching to speed up CI build

### LINKED ISSUE #82: Try to speed up CI build
The CI build is slow, mostly because the Windows build takes pretty long. While this is not a critical issue, it would be great if the build was a lot faster.

It should be possible to cache the Cargo output, so all dependencies don't have to be rebuilt all the time. I haven't done that before with GitHub Actions, but there are probably existing solutions, and it should make the build much faster.

mxdamien:
Closes #82

### DIFF
--- a/.github/workflows/fornjot.yml
+++ b/.github/workflows/fornjot.yml
@@ -16,6 +16,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @mxdamien!

I don't know a whole lot about GitHub Actions, but this looks good to me. Merging.

PR #131, opened 2022-02-04
Replace delaunator-rs with Spade

hannobraun:
This is required to clean up triangulation and address #78, as Spade supports arbitrary vertex types. This allows us to use a specialized 2D vertex type that can be converted back into a 3D point without loss due to floating point precision.


hannobraun:
@Stoeoef Thank you for Spade! This replacement was totally painless and just took a few minutes.

Issue #57, opened 2022-01-20
Provide Linux/Mac/Windows binaries


hannobraun:
To make Fornjot more accessible, and testing easier, we should compile executables for Linux, Mac, and Windows. Since we're already using GitHub Actions for the CI build, it makes sense to use it for this too. This should probably be a separate action that is triggered whenever a normal build on the `main` branch finished. It shouldn't block pull requests.

In the past, I've used the Releases feature to host those binaries. I'd rather reserve Releases for real releases though (as to not confuse users with a million different release, when they click there), so either those binaries should be downloadable through GitHub Actions, or there should be a single nightly/in-progress/WIP release that gets overwritten every time. I don't know if either of those are possible.

Labeling this as `good first issue`, as this only requires knowledge of Cargo and GitHub, not really of anything Fornjot-specific.

hendrikmaus:
FYI I am working on this in https://github.com/hannobraun/Fornjot/pull/175

hannobraun:
This is mostly done. (Thanks, @hendrikmaus!) There's one remaining issue: https://github.com/hannobraun/Fornjot/pull/175#pullrequestreview-880472207

hendrikmaus:
@hannobraun I think we can consider this one done now.

hannobraun:
Yes, thank you!

Issue #56, opened 2022-01-20
Make sure Fornjot can be used outside this repository


hannobraun:
I always just use `cargo run` during development. To make Fornjot accessible to end-users, we need to make sure that a Fornjot binary can be used outside this repository.

The one issue I'm aware of right now is that Fornjot makes assumptions about where to find models (see #9). I don't know what hidden dependencies there are. Maybe there are none.

I'm labeling this as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, since it's pretty straight-forward: Download a binary from the [Compile Binaries action](https://github.com/hannobraun/Fornjot/actions/workflows/binaries.yml) run it outside of this repository, open issues for any problems that are encountered.

hannobraun:
There are now pre-compiled binaries that could help with the testing required here. Updated the original issue description.

hannobraun:
I'm working on this.

PR #175, opened 2022-02-11
Add workflow to compile binaries for all major OS<>arch combinations

hendrikmaus:
This change-set introduces a new workflow to compile Fornjot's host application binaries for all major operating system and architecture combinations.

As defined in #57, the workflow only runs on changes to the repository `main` branch. It does not run on or block pull-requests.

Each compiled binary is compressed (unix *.tar.gz, win *.zip) and uploaded as build artifact. The target OS and architecture are used to name the archive, while the binary inside is not renamed. These build artifacts can then be downloaded, expanded and tested as one sees fit. Please mind that the artifacts only show up once all branches of the matrix build have finished.

Currently, the workflow is untested. Testing GitHub Action workflows is best done by allowing the workflow to run on the current HEAD (e.g. on this pull-request). Once asserted, the run conditions can be changed back before merging.

~@hannobraun you'd need to allow the workflow to run, because fork workflows do not run without permission.~

~Another workaround is to submit a pull-request to my own fork, which I am doing here to test it https://github.com/hendrikmaus/Fornjot/pull/1~

Tested in https://github.com/hendrikmaus/Fornjot/actions/runs/1830515553 (follow the link to also see the list of build artifacts)

---

**Impressions:**

The matrix build running:
image

The list of artifacts appears once all matrix finished running: (Windows failed)
image

Cold compile times:
image

Warm compile times: (Windows also green now)
image

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -2,7 +2,7 @@ name: Binaries
 
 on:
   push:
-    branches: [ main, '#57_provide-major-os-binaries-for-every-commit-to-main' ]
+    branches: [ main ]
 
 env:
   CARGO_TERM_COLOR: always
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -56,6 +56,7 @@ jobs:
           use-cross: ${{ matrix.cross }}
 
       - name: Compress
+        shell: bash
         run: |
           cd target/${{ matrix.target }}/release
           case "${RUNNER_OS}" in
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -1,0 +1,90 @@
+name: Binaries
+
+on:
+  push:
+    branches: [ main ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+  # Name of the crate from Cargo.toml
+  # used to package up the binaries on disk
+  PROJ_NAME: fj-host
+
+jobs:
+  binaries:
+    name: Binaries
+    strategy:
+      matrix:
+        include:
+          # Supported `cross` targets:
+          #   https://github.com/rust-embedded/cross#supported-targets
+
+          # Linux targets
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @hendrikmaus, this is awesome!

The diff looks good to me, although I am, as previously stated, not an expert at this stuff. One weird thing I noticed though: When I download one of the artifacts, it's wrapped in an additional layer of `.zip`. So it's a `.zip.zip` or `.tar.gz.zip` file. I can't see anything in the code that would explain this.

Is this something you recognize and can fix easily? If not, I'm happy to merge this now. Improvements can come later. This is already a huge step in the right direction!

Other than that, everything looks great. I just tested the Linux/musl binary, and it doesn't work in exactly the way I'd expect it to (see https://github.com/hannobraun/Fornjot/issues/56).

hannobraun:
I decided I'm going to merge this right now. I'm about to submit and merge another pull requests, which means this one will need to be rebased... it's just better to not have too much work in progress at any given point.

The solution to the problem I mentioned, whether easy or not, can come in a follow-up pull request.

hendrikmaus:
I'll have a look at the double compression. I reckon that the artifacts from the previous run got cached and were included in the archives on the subsequent one. I'll submit another pull-request to add a cleanup step after uploading the artifacts.

Good catch 👍

hendrikmaus:
Actually, it is a GitHub quirk we're running into.

When downloading an artifact via the web UI, GitHub will dynamically wrap the content in a zip archive. See https://github.com/actions/upload-artifact#zipped-artifact-downloads

We could work around this by not compressing ourselves. Since the binaries aren't very big, that should be ok. 

I'd change the logic to rename the compiled binary, so that the artifact names are unique for upload and do not override one another.

What do you think about that plan @hannobraun ?

hannobraun:
Sounds good, @hendrikmaus!

PR #179, opened 2022-02-12
Iterate on workflow to compile binaries for all major OS<>arch combinations

hendrikmaus:
This change-set iterates on the initial version of the workflow https://github.com/hannobraun/Fornjot/issues/57 and fixes the behavior observed and described in https://github.com/hannobraun/Fornjot/pull/175#pullrequestreview-880472207.

Instead of compressing the compiled binaries before they are uploaded to GitHub, they are renamed to `fj-host-<target>`:
image

Furthermore, I adjusted the name of the workflow and some comments in the file.

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -59,13 +59,17 @@ jobs:
         shell: bash
         run: |
           # Include compile target in binary name
-          
-          mv -v \
-            "target/${{ matrix.target }}/release/${PROJ_NAME}" \
-            "${GITHUB_WORKSPACE}/${PROJ_NAME}-${{ matrix.target }}"
-          
-          chmod -v +x \
-            "${GITHUB_WORKSPACE}/${PROJ_NAME}-${{ matrix.target }}"
+
+          src="target/${{ matrix.target }}/release/${PROJ_NAME}"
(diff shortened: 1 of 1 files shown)


hannobraun (commented):
Thank you, @hendrikmaus!

The commits look good, but we have a new problem: The execute permissions don't survive the upload/download process, without an archive to wrap them. (At least for the Linux/musl binary, but I suspect it's true for all the non-Windows ones.)

I'm tempted to just merge, as this pull request contains other improvements. Not sure if missing permissions are any better than the double-archive. Not much of a difference to me personally, but I suspect a less experienced user might be less confused by the double-archive.

Any suggestions?

hendrikmaus:
I tried to explicitly retain the executable flag, but looks like we're out of luck.

On releases, we won't have these issues though. I reckon new users will either download the binary from a release or try to use a package manager to install. Hence, these "intermediary" binaries will only be used by contributors.

I'd stick with the current solution. However, we should write a little piece of documentation.

hannobraun:
Makes sense, thank Hendrik! I'm merging now, and will open an issue about documentation shortly.

hannobraun:
Opened #180.

Issue #139, opened 2022-02-07
Add Clippy to CI build


hannobraun:
I think it makes sense to add Clippy to the CI build, as it often has valid complaints. However, it shouldn't be added blindly. First, we shouldn't add it, if it causes a million warnings right away. There should be a clean-up first, to make sure all applicable warnings are fixed. Second, if during that clean-up we come across Clippy warnings that aren't applicable, those should be disabled.

Labeling as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this doesn't require a lot of knowledge of Fornjot, and going through the warnings could be a good way to get to know the code base a little.

ObiWanRohan:
Is someone working on this?

hannobraun:
Not to my knowledge. @hendrikmaus was working on adjacent build system topics, but he said he's interested in #104.

Feel free to pick this up, if you're interested! Please note that there's some discussion about how to add more things to the CI build in https://github.com/hannobraun/Fornjot/issues/64 that is relevant here too.

hendrikmaus:
I can recommend to use this step: https://github.com/hendrikmaus/rust-workflows/blob/master/.github/workflows/ci.yaml#L90

Clippy should not run in a matrix, linting the codebase on one OS is enough.

hendrikmaus:
Basically, we can use the entire workflow here as is https://github.com/hendrikmaus/rust-workflows/blob/master/.github/workflows/ci.yaml

hannobraun:
> Clippy should not run in a matrix, linting the codebase on one OS is enough.

Thank you, @hendrikmaus, good point!

I still want a script with all the essentials (`test`, `clippy`, `fmt`, ...) that a developer can execute locally, to gain confidence that CI will pass. Going back-and-forth with the CI build can be a nuisance, especially if you're new to a project and don't know everything that the CI build checks for.

That doesn't have to be a part of this issue though. I'd be happy to call this complete, if we have warning-free Clippy build that is checked on CI.

hendrikmaus:
Regarding local checking, I usually run multiple terminals + cargo watch to run clippy and tests on every code change. Those commands are either in my shell history or I have a Justfile or Makefile wrapping them.

One could also look into making that setup easier with https://github.com/Canop/bacon which recently released a brand new major version.

hannobraun:
Interesting, I didn't know bacon!

[...]

PR #206, opened 2022-02-18
build: add `rust-toolchain.toml` configuration file

hendrikmaus:
This change-set introduces the `rust-toolchain.toml` file to the root of the project. This file will ensure that all contributors and CI pipelines always compile the project with the same Rust version.

This file is automatically picked up by `rustup`, which is the defacto standard in the Rust ecosystem to manage the toolchain.

You can test and see the override by checking out this branch and running:

_Aside: in the example below, my system already had the toolchain installed. If that is not the case, see below._

```shell
$ rustup toolchain list
stable-x86_64-unknown-linux-gnu (default)
nightly-x86_64-unknown-linux-gnu
1.58.1-x86_64-unknown-linux-gnu (override)
```

And you happen to see `1.58.1-x86_64-unknown-linux-gnu (override)`, where `override` indicates that `rustup` read the configuration and uses the desired toolchain.

If you do not (yet) have the desired toolchain installed, the first call to build the project using `cargo` will automatically leverage `rustup` to install the toolchain. You can try by:

```shell
$ rustup uninstall 1.58.1-x86_64-unknown-linux-gnu

$ rustup toolchain list
stable-x86_64-unknown-linux-gnu (default)
nightly-x86_64-unknown-linux-gnu

$ cargo build
info: syncing channel updates for '1.58.1-x86_64-unknown-linux-gnu'
info: latest update on 2022-01-20, rust version 1.58.1 (db9d1b20b 2022-01-20)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
```

PR #208, opened 2022-02-18
Add clippy to CI and fix warnings

ObiWanRohan:
Add a workflow to run Clippy.

Solving issue #139


.github/workflows/test.yml
```diff
@@ -0,0 +1,102 @@
...
+      - main
+      - master
+    paths-ignore:
+      - "docs/**"
+      - "**.md"
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
```
hannobraun, line 21: What does this mean, in practical terms? Does it leave out draft pull requests?

cc @hendrikmaus

.github/workflows/test.yml
```diff
@@ -0,0 +1,102 @@
...
+# Continuous Integration Workflow For Rust
+name: CI
+
+# Define the triggers; usually merges to the repository
+# default branch and pull-requests
+on:
+  push:
+    branches:
+      - main
+      - master
```
hannobraun, line 13: We actually just need the `main` branch here.
```suggestion
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,102 @@
...
+
+  clippy:
+    name: Clippy Check
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+          - macOS-latest
+          - windows-latest
+    runs-on: ${{matrix.os}}
```
hannobraun, line 91: As per the discussion in #139, I don't think we need to run Clippy on all operating systems.
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,102 @@
...
+  CARGO_TERM_COLOR: always
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-20.04
```
hannobraun, line 41: What's the reasoning for using `ubuntu-20.04` here? `ubuntu-latest` is used below.

cc @hendrikmaus

hannobraun (changes requested):
Thank you for the pull request, @ObiWanRohan!

I've left some comments with questions and change suggestions.

In addition, there are some problem and questions that need to be addressed before this can be merged:

- The new jobs are failing, complaining about a missing rust-toolchain file. I don't understand why. One was added earlier today (https://github.com/hannobraun/Fornjot/pull/206), and this pull request seems to be up-to-date with `main`.
- Does it make sense to also run `cargo build`, not just `cargo test`? They're *mostly* redundant, but a `cargo build` might expose additional problems, like linker issues. Might be irrelevant right now.
- We already have [existing CI jobs](https://github.com/hannobraun/Fornjot/blob/main/.github/workflows/fornjot.yml) that would need to be removed, or merged into the new jobs.
- I'm concerned about using so many different jobs. Wouldn't starting a new runner cause overhead each time, resulting in a longer CI build? Wouldn't it be better to just have one job per platform, with the appropriate steps specified for each? That would also cut down on some redundant `uses` declarations.
- Can we configure Dependabot to update all those actions? It might be tedious to keep them up-to-date manually.

ObiWanRohan:
I should've created this as a draft PR.
This was just an initial commit to have a baseline of what was happening so bear with me as I ready this PR.

About your questions:
- About the rust toolchain issue, this seems to be an issue with the action. It isn't reading the new `rust-toolchain.toml` file. 
The issue is being tracked [here](https://github.com/actions-rs/toolchain/issues/126).
**For a workaround** - Is there a specific reason to use Rust toolchain version `1.58.1`? Or can we pin it to stable? It can be used as a static argument to the workflow.
- For the second and third points, I thought this workflow would be separate than the other workflows, and this one would run more frequently than the other build workflows.
- This workflow is only going to run on one platform, so I'm not sure if it would matter currently
- I'll check if Dependabot can do it. It would be of much help if it has the ability.

hannobraun:
> I should've created this as a draft PR.
> This was just an initial commit to have a baseline of what was happening so bear with me as I ready this PR.

Understood. I've converted the pull request to a draft (I think you could have done this too; there's a subtle link in the upper right under *Reviewers*). Just convert it back, once you feel it's ready.



> * About the rust toolchain issue, this seems to be an issue with the action. It isn't reading the new `rust-toolchain.toml` file.
>       The issue is being tracked [here](https://github.com/actions-rs/toolchain/issues/126).
>       **For a workaround** - Is there a specific reason to use Rust toolchain version `1.58.1`? Or can we pin it to stable? It can be used as a static argument to the workflow.

Ah, got it. Yeah, I think we can pin it to stable, as a workaround. I don't feel strongly anyway, about whether we should pin to stable in `rust-toolchain.toml`, or whether we should control the version tightly.


> * For the second and third points, I thought this workflow would be separate than the other workflows, and this one would run more frequently than the other build workflows.
> * This workflow is only going to run on one platform, so I'm not sure if it would matter currently

The workflow I linked runs on every push to a pull request, so just as often as this one. There's really no difference between them, except that this one is new and contains more stuff. There's the same intent behind them, to test pull requests.

There's also the [*Compile Binaries* workflow](https://github.com/hannobraun/Fornjot/blob/main/.github/workflows/binaries.yml) (might get renamed soon), which only runs when `main` is pushed to. That one is separate.



> * I'll check if Dependabot can do it. It would be of much help if it has the ability.
> Looks like Dependabot has the ability to update the actions workflows too

That's great, thanks for checking :+1:

.github/workflows/test.yml
```diff
@@ -0,0 +1,95 @@
...
+      - synchronize
+    paths-ignore:
+      - "docs/**"
+      - "**.md"
+  # The following makes this workflow re-usable by other repositories
+  # without having to copy the entire thing.
+  #
+  # See https://docs.github.com/en/actions/learn-github-actions/reusing-workflows
+  workflow_call: {}
+
```
hendrikmaus, line 29:
```suggestion

```

This makes the original workflow be a re-usable one. We can drop that here.

hendrikmaus:
> About the rust toolchain issue, this seems to be an issue with the action. It isn't reading the new rust-toolchain.toml file.
The issue is being tracked https://github.com/actions-rs/toolchain/issues/126.
For a workaround - Is there a specific reason to use Rust toolchain version 1.58.1? Or can we pin it to stable? It can be used as a static argument to the workflow.

That is a really unfortunate issue. I did not see that one coming 🤔 I think pinning to stable is fine though.

> I'm concerned about using so many different jobs. Wouldn't starting a new runner cause overhead each time, resulting in a longer CI build? Wouldn't it be better to just have one job per platform, with the appropriate steps specified for each? That would also cut down on some redundant uses declarations.

@hannobraun since jobs run in parallel, the time issue is negligible. The matrix job should be kept separate, but the rest can be combined if you are more comfortable with it. Originally, I put them into individual ones for readability in the checks section of a pull-request. At a glance, you'd immediately see what part of CI went south.

> The workflow I linked runs on every push to a pull request, so just as often as this one. There's really no difference between them, except that this one is new and contains more stuff. There's the same intent behind them, to test pull requests.

The existing workflow https://github.com/hannobraun/Fornjot/blob/main/.github/workflows/fornjot.yml is superseded by this one can can be removed.

---

Glad to get some feedback out of this to improve rust-workflows :)

hendrikmaus:
I also added the _same_ `rust-roolchain.toml` to rust-workflows and everything builds flawlessly 🤔 https://github.com/hendrikmaus/rust-workflows/pull/40

hendrikmaus:
> You're definitely the expert here. So if you're telling me it's best practice to do it this way, and it won't affect the runtime negatively (at least most of the time), that's fine by me.

Thanks for the praise. I'd stick with this job-split until we notice a negative impact that is significant enough to hurt the workflow. Until then, I'd keep it like it is in favor of the user experience when glancing at the "checks" section of a pull-request.

.github/workflows/test.yml
```diff
@@ -0,0 +1,95 @@
...
+          command: test
+
+  clippy:
+    name: Clippy Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+      - uses: actions-rs/toolchain@568dc894a7f9e32ffd9bb7d7a6cebb784cdaa2b0
+        with:
+          profile: minimal
```
hendrikmaus, line 89:
```suggestion
          toolchain: stable
          profile: minimal
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,95 @@
...
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+          - macOS-latest
+          - windows-latest
+    runs-on: ${{matrix.os}}
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+      - uses: actions-rs/toolchain@568dc894a7f9e32ffd9bb7d7a6cebb784cdaa2b0
```
hendrikmaus, line 76:
```suggestion
      - uses: actions-rs/toolchain@568dc894a7f9e32ffd9bb7d7a6cebb784cdaa2b0
        with:
          toolchain: stable
```

hendrikmaus:
I added a couple of suggestions with the stable toolchain. That way, it also works as expected in rust-workflows.

.github/workflows/test.yml
```diff
@@ -0,0 +1,95 @@
...
+        with:
+          profile: minimal
+      - uses: Swatinem/rust-cache@1232abb8968faf344409165de17cbf9e7f340fd8
+      - uses: actions-rs/cargo@4ff6ec2846f6e7217c1a9b0b503506665f134c4b
+        with:
+          command: check
+
```
hendrikmaus, line 50:
```suggestion
```

Actually, I cam to the conclusion that we can drop `cargo check` because we are running `cargo clippy`. The latter does the same as the former and supersedes it with more functionality.

hannobraun: Yeah, makes sense.

ObiWanRohan:
> I also added the same rust-roolchain.toml to rust-workflows and everything builds flawlessly 🤔 https://github.com/hendrikmaus/rust-workflows/pull/40

@hendrikmaus You're correct. I added the `override: stable` argument to the toolchain action, and the `rust-toolchain.toml` file was picked by `rustup`.

hendrikmaus:
Ci-wise this looks fine now @ObiWanRohan - thank you.

---
The changes in the `src` directory shouldn't show up in this change-set though 🤔

ObiWanRohan:
> The changes in the src directory shouldn't show up in this change-set though 🤔

Those are due to the trivial fixes which I did from the Clippy suggestions and the merge from the `main` branch to resolve conflicts

hendrikmaus:
Ah got it, should've looked at the change-set myself 🤦🏻 

IMHO I would always submit two or more change-sets in this situation to keep the concerns separated: 1) mutate the CI pipeline 2) code changes. That is because either change can have side-effects which could make you revert partial patches. That is way cleaner if each change-set is bound to a single concern. If that makes sense :)

@hannobraun are you happy to merge the change-set like this anyway?

hannobraun:
I also prefer to have separate concerns (like fixing Clippy warnings and making CI changes) in separate pull requests. One additional reason, on top of what @hendrikmaus mentioned, is to get changes in as soon as possible. I might merge a change today that conflicts with Clippy fixes in this PR, which would then cause unnecessary work for @ObiWanRohan.

As a general rule, everything that makes sense on its own should be merged right away (but obviously you don't need to make one pull request per minute for 10 minutes, if you're doing a bunch of small cleanups; use your judgement).

But that's not a blocker. I'd be happy to merge the PR anyway.

There are two things I'd like to see done before merging, ideally:
- Fix the existing Clippy warnings, ideally in a separate pull request. I'd prefer the check to be green when merging it.
- Remove the now redundant "Build Fornjot" workflow.

But those aren't hard blockers either. If @ObiWanRohan ran out of time and couldn't work on this for a while, for example, I'd prefer to merge the existing improvements and make further changes later, instead of letting this sit here to suffer from bit rot.

Also, as a general note, if merge conflicts need to be resolved in a pull request, I prefer the pull request to be rebased, instead of a merge commit being created. That leaves a simpler (and thus easier to review) history. This isn't a hard requirement either.

(And by the way, this isn't a full review. I haven't looked at the commits for a while; this is just some stuff I noticed.)

therealprof:
Whoops, sorry, I just noticed this PR. I've fixed some of the lints with #232. My bad.

The now remaining lints were not quite as clear cut which why I haven't addressed them yet. Might require some more consideration...

ObiWanRohan:
I just got back to this PR.

I agree that the Clippy fixes are probably better suited to a separate PR.
 @hannobraun Should I revert the commits with the Clippy changes or is it okay if I reset my branch and do a force push?

hannobraun:
@therealprof 
> Whoops, sorry, I just noticed this PR. I've fixed some of the lints with #232. My bad.

My preferred solution to that problem is to get changes merged as early as practical, exactly so that contributors like you don't have to be aware of all ongoing work :smile: 


@ObiWanRohan 
> Should I revert the commits with the Clippy changes or is it okay if I reset my branch and do a force push?

My favorite solution would be to have the Clippy fixes in a separate pull request (or even multiple, if some are uncontroversial and can be merged right away, while other require discussion), then rebase/force-push this pull request to have the CI changes as a separate pull request with a clean history.

But the whole point of my preferences is to make the workflows here efficient, and that's going to depend on the developer and the work. If you think you might want to fiddle with the CI some more, it's probably best to submit the other changes separately. If you want to do a final push to get everything over the finish line in one go, than just do that. Whatever works best for you!

ObiWanRohan:
I have removed the previous checks (`LinuxLatest`, `MacOsLatest`, and `WindowsLatest`), and added the same steps to the new workflow. Those would run in the matrix check.

.github/workflows/test.yml
```diff
@@ -0,0 +1,84 @@
...
+      - uses: actions-rs/toolchain@568dc894a7f9e32ffd9bb7d7a6cebb784cdaa2b0
+        with:
+          toolchain: stable
+          profile: minimal
+      - run: rustup component add clippy
+      - uses: Swatinem/rust-cache@1232abb8968faf344409165de17cbf9e7f340fd8
+      - uses: actions-rs/clippy-check@9d09632661e31982c0be8af59aeb96b680641bc4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --all-features
```
hendrikmaus, line 84:
```suggestion
          args: --all-features
          name: CI / Clippy run
```
This should make for a nicer name to appear in the "checks" section of a pull-request. By default, the job's name appears as `clippy_check`.

hendrikmaus:
I think this one is ready to be merged 👍🏻 Nicely done @ObiWanRohan

.github/dependabot.yml
```diff
@@ -4,3 +4,9 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every weekday
```
hannobraun, line 11: That comment doesn't match the code. I think weekly is fine.
```suggestion
```

hannobraun (approved):
I agree that this looks very good. Thank you @ObiWanRohan!

I'm happy to merge as-is, with a few reservations:
- I found one tiny issue. Left a comment. I can probably apply the fix myself, after leaving this review.
- I'd prefer to have all Clippy warnings fixed or suppressed before merging, but it's also okay to fix them soon-ish after merging this check. I'll open an issue.
- GitHub tells me this branch is out-of-date with the base branch, but doesn't show me the button to let me rebase it. I'll look into that after leaving the review.

If I can cajole GitHub into letting me rebase, I'll merge this right after leaving this review. If I have to manually rebase, I'll do that later today.

Thanks again, @ObiWanRohan, and thank you @hendrikmaus for the help!

hannobraun (changes requested):
Ah, I can't apply my own review suggestion. Looks like @ObiWanRohan didn't check that box that would allow me to push changes to the branch. This would also explain why I can't rebase.

So, happy to merge, once:
1. My suggestion is applied.
2. @hendrikmaus's suggestion also looks good. I see no reason not to apply that too.
3. The branch has been rebase on `main`.

.github/workflows/test.yml
ObiWanRohan: The name currently appears as `CI / Clippy Check`, the name is configured here

https://github.com/ObiWanRohan/Fornjot/blob/6155a28d9ee469440edb479134b9a87c1396ba5f/.github/workflows/test.yml#L71

ObiWanRohan:
@hannobraun The PR is ready to merge. 
There are some workflows which are shown as the 3 expected checks. I think that is something to be done with the merge settings

hannobraun (approved):
Thank you, @ObiWanRohan, yet again!

> There are some workflows which are shown as the 3 expected checks. I think that is something to be done with the merge settings

Yes, those must be because of the branch protection settings. I should be able to take care of this on my end.

hannobraun:
Opened #261.

.github/workflows/test.yml
hendrikmaus: The job emits a check and the step it runs emits another. I don't know a way to solve that right now.

hannobraun: I don't think it's much of a problem. Just mentioned it because it seemed related to the discussion here regarding the name displayed under "checks".

PR #217, opened 2022-02-19
Add a hacky model bounding box info to rendering UI

### LINKED ISSUE #134: Adding dimension rendering?
I was having a quick look at how difficult it would be to add an overlay to display the dimensions of the rendered object but I got lost in the tons of indirections to figure out where the actually rendered data would have to be created.

therealprof:
Closes #134

### DIFF
--- a/src/graphics/config_ui.rs
+++ b/src/graphics/config_ui.rs
@@ -6,6 +6,8 @@ use wgpu_glyph::{
     GlyphBrush, GlyphBrushBuilder, Section, Text,
 };
 
+use crate::math::Aabb;
+
 use super::{draw_config::DrawConfig, COLOR_FORMAT};
 
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


therealprof:
@hannobraun As you mentioned the current model rendering and UI code is a bit limited so maybe this would suffice as temporary hack.

hannobraun (approved):
Thank you, @therealprof, looks good!

> As you mentioned the current model rendering and UI code is a bit limited so maybe this would suffice as temporary hack.

Yes, definitely. No point in trying to have the perfect UI, based on the current infrastructure :smile:

therealprof:
Whoops, I noticed I forgot to add a ChangeLog entry. 😅

hannobraun:
> Whoops, I noticed I forgot to add a ChangeLog entry. :sweat_smile:

That's actually not necessary! I prefer to update the changelog at release time. That puts less of a burden on contributors and is not much extra work for me, since I edit the changelog pretty heavily anyways (by grouping related work into a single entry, for example).

PR #218, opened 2022-02-20
Fixed example in README to actually work

therealprof:


hannobraun (approved):
Thank you, @therealprof, good catch!

PR #233, opened 2022-02-22
build: introduce bacon configuration file

hendrikmaus:
_Aside: it is perfectly fine to reject this pull-request, if the root of the project gets too many tool-specific config files._

This change-set introduces a project-level configuration for the developer utility [`bacon`](https://dystroy.org/bacon/). Bacon allows for running various jobs, e.g. `cargo check` continuously. It functions similarly to `cargo watch` (also a 3rd party plugin to `cargo`) in that it monitors the filesystem and re-runs the given job at hand. One can run multiple instances of bacon in multiple terminal windows. There are no known side-effects on other tools that I know of.

I created the default configuration using `bacon --init`. Then I added a job called [`nextest`](https://nexte.st/) (short cut `n`) to use `nextest` (`cargo install nextest`; `cargo nextest run`), which is quite an interesting new test runner that I am starting to use across projects.

I tried to use bacon's global configuration to define jobs, but that is not supported. Jobs are project-level only. However, one can set certain defaults for oneself (see https://dystroy.org/bacon/#global-preferences) if desired.

I left the default job at `cargo check`, but we can change it to `clippy` later on.

### DIFF
--- a/bacon.toml
+++ b/bacon.toml
@@ -42,7 +42,7 @@ need_stdout = false
 [jobs.doc-open]
 command = ["cargo", "doc", "--color", "always", "--no-deps", "--open"]
 need_stdout = false
-on_success = "back"                                                    # so that we don't open the browser at each change
+on_success = "back" # so that we don't open the browser at each change
 
 # You can run your application and have the result displayed in bacon,
 # *if* it makes sense for this crate. You can run an example the same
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thanks, @hendrikmaus!

Happy to merge this, and looking forward to trying Bacon and nextest.

I think having this kind of configuration in the repository (along with other configuration, like IDE-specific settings) is fine, as long as they get used. If we accumulate too much stuff, I might clean out configuration that doesn't get used contributors regularly.

PR #238, opened 2022-02-22
Privatize Triangle struct

therealprof:


hannobraun (approved):
Thank you, @therealprof!

This makes sense. It will allow us to validate triangles in the future (make sure these are three different points not on a line).

PR #239, opened 2022-02-22
Privatize Segment struct

therealprof:

### DIFF
--- a/src/math/segment.rs
+++ b/src/math/segment.rs
@@ -5,8 +5,8 @@ use super::Point;
 /// A line segment, defined by its two end points
 #[derive(Clone, Copy, Eq, PartialEq, Hash, Ord, PartialOrd)]
 pub struct Segment<const D: usize> {
-    pub a: Point<D>,
-    pub b: Point<D>,
+    a: Point<D>,
+    b: Point<D>,
 }
 
 impl Segment<2> {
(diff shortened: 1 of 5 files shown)


src/math/segment.rs
```diff
@@ -5,22 +5,30 @@ use super::Point;
...
 
 impl Segment<2> {
     /// Convert the 2-dimensional segment to a Parry segment
     pub fn to_parry(&self) -> parry2d_f64::shape::Segment {
         [self.a.to_na(), self.b.to_na()].into()
     }
+
+    pub fn points(&self) -> [Point<2>; 2] {
+        [self.a, self.b]
+    }
```
hannobraun, line 20: I think this method can be unified with the identical one below, by putting it in an `impl<const D: usize> for Segment<D> { ... }` block and returning `[Point<D>; 2]`.

hannobraun (approved):
Thank you, @therealprof!

Like #238, this makes sense to me, as it lets us validate segments (make sure the points are not identical) as a next step.

I left a comment with a possible improvement, but that's not critical. I'm going to merge this PR right away. Feel free to make the improvement in a follow-up PR, if you want, or leave it to someone else to do later :grin:

src/math/segment.rs
hannobraun: I was working on the math code anyway, so I quickly did this in #243.

PR #249, opened 2022-02-25
Implement triangle validation and add unit tests

### LINKED ISSUE #240: Validate `math::Triangle` on construction
[`math::Triangle`](https://github.com/hannobraun/Fornjot/blob/dfc34e74b4262bfd3edf5f13ef0de821e229048c/src/math/triangle.rs) should be validated on construction. If buggy code constructs an invalid triangle, that should cause an immediate panic, so the bug can be discovered and fixed as easily as possible.

In specific terms, validation includes the following steps:
- Checking that the three points are distinct.
- Checking that the three points don't lie on the same line.

See [`math::Segment`](https://github.com/hannobraun/Fornjot/blob/a774cc94fb063d062074fee68a8e3a1f4c1c078e/src/math/segment.rs#L33-L46) for an example of this kind of validation.

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as those are relatively small changes that are very localized.

therealprof:
Closes #240

### DIFF
--- a/src/math/triangle.rs
+++ b/src/math/triangle.rs
@@ -1,4 +1,4 @@
-use super::Point;
+use super::{Point, Scalar};
 
 /// A triangle
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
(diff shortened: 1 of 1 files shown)


therealprof:
Please note that for some reason I can't seem to be able to create a 2D triangle because cross product somehow only works for 3D vertices?

hannobraun (approved):
Thank you, @therealprof, looks good!

> Please note that for some reason I can't seem to be able to create a 2D triangle because cross product somehow only works for 3D vertices?

First, I'm going to merge this besides this limitation. The math code isn't exposed in a public library (yet), and we don't seem to create 2D triangles, so it's fine for now.

Second, I think I get it:
- Mathematically, at least according to my primitive understanding, the cross product is not defined for 2D vectors. The result is a vector that is perpendicular to the input vectors, which can't work in 2D.
- So the question becomes, why does `Vector::cross` exist for 2D vectors? Answer: It was an accident, made possible because `Matrix::cross` exists for all matrices in nalgebra.
- Digging into that, [`nalgebra::Matrix::cross`](https://docs.rs/nalgebra/latest/nalgebra/base/struct.Matrix.html#method.cross) is documented to panic for non-3D vectors, and will be removed for non-dynamic/non-3D vectors in the future.

Idea for fixing this:
- Add a method that converts a `Vector<D>` into a `Vector<3>`. If it's a `Vector<2>`, the `z` component can be zero.
- Call that method before computing the triangle area, so the cross product is always done on 3D vectors.

In addition, we should define `Vector::cross` only for `Vector<3>`.

I've made notes about all of these action items. I'll either do them myself later, or open issues to track them.

therealprof:
You're right. The cross product is a 3D of the generalised version of the external product. In 2D the cross product degenerates to the determinant. I think the reason why `nalgebra` only allows for 3-dimensional vectors is because only then the result will be again a vector. A long time ago I might have known this... 😅

Issue #280, opened 2022-03-03
Make `Shape` consistent by design


hannobraun:
[`Shape`](https://github.com/hannobraun/Fornjot/blob/370bc715e54504ae6a7448835635a8be005f00e2/src/kernel/topology/mod.rs#L11-L43) currently does not do a lot to make sure that the shape it represents is consistent in any way. Its internal representation is redundant and can largely be manipulated directly by callers, which is error-prone.

`Shape` wasn't designed to be that way. It has grown from separate structs in https://github.com/hannobraun/Fornjot/pull/254. I've since made some strides to improve this, by locking down the vertex aspect of it (https://github.com/hannobraun/Fornjot/pull/276), but more such work remains to be done.

The end goal for `Shape` should be something like the following:
- There should be no internal redundancy. Geometric and topological data should be stored in a central place, and referred to where it's needed.
- It should always represent a consistent shape, as far as that is practical. For example, it shouldn't be possible to remove a vertex, without first removing the edges that refer to it.
- Its public API should enforce these constraints.
- Furthermore, the API should make manipulation shapes straight-forward.

When making these changes, I think [`transform`](https://github.com/hannobraun/Fornjot/blob/370bc715e54504ae6a7448835635a8be005f00e2/src/kernel/algorithms/transform.rs) and [`sweep`](https://github.com/hannobraun/Fornjot/blob/370bc715e54504ae6a7448835635a8be005f00e2/src/kernel/algorithms/sweep.rs) can be useful guide posts. They are messy, to a large degree because `Shape` is messy.

hannobraun:
I'm going to start working on this. See https://github.com/hannobraun/Fornjot/issues/242#issuecomment-1058190651.

PR #288, opened 2022-03-04
Refer to vertices by handle

hannobraun:
This is a step towards #280. It adds a `Handle` type that can be used to refer to geometric or topological objects by reference instead of by value, and updates `Vertex` to do that.

This doesn't make a difference yet. The infrastructure is put in place, but not used for anything new yet. It's an important step towards removing the redundancy in `Shape` though, as the new infrastructure makes it possible to update the *value* of a vertex, without having to update all references to it. The latter has turned out to be not only tedious, but very error-prone.


hannobraun:
This pull request is ready to merge now. I've updated the title and description to reflect its new scope.

PR #292, opened 2022-03-06
address a few lints

danieleades:


hannobraun (approved):
Thank you for the pull request, @danieleades. Looks great!

PR #337, opened 2022-03-12
Start with a larger glyph cache size from the get go

therealprof:
This eliminates the startup warning

```2022-03-12T13:07:34.975288Z  WARN wgpu_glyph: Increasing glyph texture size (256, 256) -> (512, 512). Consider building with `.initial_cache_size((512, 512))` to avoid resizing```


hannobraun (approved):
Thank you, @therealprof

PR #338, opened 2022-03-12
Allow Triangles to carry a color and transfer that to the vertices du…

therealprof:
…ring meshing

This moves the hardcoded color specification (still red with full
opacity) one level closer to the actual model.


hannobraun (approved):
Thank you, @therealprof!

I think adding the color to `Triangle` itself feels a bit icky, as color seems like a separate concern that the `math` types shouldn't have to care about. Nonetheless, this is a practical solution, and good enough for now (my favorite kind of good).

I'm going to merge it, especially since there's already a PR for adding coloring support based on it #343. We can clean this up later, if and when someone comes up with something better.

PR #343, opened 2022-03-13
Coloring support

### LINKED ISSUE #291: Colouring support
It would be great if Fornjot would have support for specifying custom colors per object in the rendering rather than the default of red with full opacity. This might also be useful to figure out which specific item is causing rendering issues.

There's a number of issues to sort out though:

- Which type to use? I'm not super keen on using the current `[f32;4]` everywhere since it doubles up the issues FJ already has with the floating point representation of coordinates and would propose [u8; 4] instead.
- Currently vertices are compared for equality, adding a colour to a vertex would not only mean more iffiness with the passthrough of the color information but also create ambiguities with respect to "is an otherwise identical Vertex with a different colour a different Vertex or the same"?
- I suppose a `Colour` type would live in it's own module to allow convenient use and manipulation. What's the best location for it?

therealprof:
Implement basic coloring support for shapes

The examples have been extended to demonstrate the coloring of objects. A few bits and pieces are probably still missing or could be improved but let's look at those in another step.

Closes #291

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,8 +54,8 @@ pub extern "C" fn model(args: &HashMap<String, String>) -> fj::Shape {
         .parse()
         .unwrap();
 
-    let outer_edge = fj::Circle { radius: outer };
-    let inner_edge = fj::Circle { radius: inner };
+    let outer_edge = fj::Circle::from_radius(outer);
+    let inner_edge = fj::Circle::from_radius(inner);
 
     let footprint = fj::Difference {
         a: outer_edge.into(),
--- a/README.md
+++ b/README.md
@@ -57,10 +57,7 @@ pub extern "C" fn model(args: &HashMap<String, String>) -> fj::Shape {
     let outer_edge = fj::Circle::from_radius(outer);
     let inner_edge = fj::Circle::from_radius(inner);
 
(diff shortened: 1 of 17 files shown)


hannobraun (approved):
Thank you, @therealprof!

I think this looks good, overall. There are some things that could be improved, but as you state in your last commit, let's look at those later.

Some thoughts on improvements:
1. There should be a `Color` type that wraps the `[u8; 4]` and provides a nice API around it.
2. I don't like how color is added to the various `fj` types.

Regarding 2., not sure what would be better. Maybe an explicit `fj::Color` operation, that colors everything it wraps (might not be flexible enough, for example regarding different colors for sweep side walls, as your comment mentions).

Or maybe there could be a type `Object<T>` that would wrap `Shape2d`/`Shape3d`/`Shape` everywhere those are used, and add color information (and in the future, potentially other information that all objects share).

But that doesn't really matter right now, I'm just throwing out ideas. I'll be merging this pull request shortly.

hannobraun:
Hmm, GitHub wouldn't let me rebase, claiming there were conflicts, but doing so manually was no problem. Will merge as soon as CI is green.

PR #352, opened 2022-03-15
Add command line argument for user defined model tolerance

### LINKED ISSUE #118: Make tolerance overridable
Fornjot works with a tolerance value, that specifies how far a generated triangle mesh is allowed to deviate from the ideal geometry that a model defines. Right now, this value is [computed automatically](https://github.com/hannobraun/Fornjot/blob/f702a4193a28899899ce989097737d462f382862/src/main.rs#L68-L84), based on the total size of the model. This is fine as a fallback, but the user should be able to override this value.

A command-line argument can be added in [`args.rs`](https://github.com/hannobraun/Fornjot/blob/main/src/args.rs) and then processed in [`main.rs`](https://github.com/hannobraun/Fornjot/blob/main/src/main.rs). As of this writing, the tolerance is also computed in `main.rs`, so integrating an argument there should be straight-forward. Labeling as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this is a small change that doesn't require much knowledge of Fornjot.

mxdamien:
closes #118

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -14,6 +14,10 @@ pub struct Args {
     /// Parameters for the model, each in the form `key=value`
     #[clap(short, long)]
     pub parameters: Vec<String>,
+
+    // Model deviation tolerance
+    #[clap[short, long]]
+    pub tolerance: Option<f64>,
 }
 
 impl Args {
--- a/src/main.rs
+++ b/src/main.rs
@@ -81,22 +81,32 @@ fn main() -> anyhow::Result<()> {
 
     let mut aabb = shape.bounding_volume();
 
-    // Compute a reasonable default for the tolerance value. To do this, we just
-    // look at the smallest non-zero extent of the bounding box and divide that
-    // by some value.
-    let tolerance = {
-        let mut min_extent = Scalar::MAX;
-        for extent in aabb.size().components() {
-            if extent > Scalar::ZERO && extent < min_extent {
-                min_extent = extent;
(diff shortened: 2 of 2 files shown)


src/args.rs
```diff
@@ -14,6 +14,10 @@ pub struct Args {
     /// Parameters for the model, each in the form `key=value`
     #[clap(short, long)]
     pub parameters: Vec<String>,
+
+    // Model deviation tolerance
+    #[clap[short, long, default_value = "0.0"]]
+    pub tolerance: f64,
```
hannobraun, line 20: I don't like how an invalid value is used as the default. I realize this is handled correctly below, but it's still weird, and also potentially misleading, as this information is used for the auto-generated command-line documentation.

How about:
```suggestion
    #[clap[short, long]]
    pub tolerance: Option<f64>,
```
I *think* that should work. Then handle the `Option` accordingly. If the user specifies an invalid value (zero or less), that can result in an error message.

I believe you should be able to call `anyhow::bail!("error message: {}", some_value)` in `main` to produce a nice error message. If that doesn't work, a panic will do for now.

hannobraun (commented):
Thank you, @mxdamien! Great to see you back!

I've left a comment with a suggestion.

src/args.rs
mxdamien: Thanks for the feedback!

hannobraun (approved):
Thank you, @mxdamien, looks and works great!

Issue #369, opened 2022-03-16
Continuous Deployment build fails


hannobraun:
See https://github.com/hannobraun/Fornjot/actions/runs/1993939647.

This is most likely due to #368. I'll take a look tomorrow.

cc @hendrikmaus

hendrikmaus:
That error message a new one for me:

```
error: cannot produce cdylib for `cuboid v0.1.0 (/project/models/cuboid)` as the target `aarch64-unknown-linux-musl` does not support these crate types
```

According to Cargo's documentation, this should work on all major operating systems:

> --crate-type=cdylib, #[crate_type = "cdylib"] - A dynamic system library will be produced. This is used when compiling a dynamic library to be loaded from another language. This output type will create *.so files on Linux, *.dylib files on macOS, and *.dll files on Windows.
>
> Source: https://doc.rust-lang.org/reference/linkage.html

I wonder if it has something to do with GNU vs musl 🤔
 
_(...)_

I think that is the culprit, I can compile the GNU targets, but musl refuses to:

```shell
$ cargo build --release --target x86_64-unknown-linux-musl
error: cannot produce cdylib for `cuboid v0.1.0 (/home/hmaus/Code/Opensource/Rust/fornjot/repo/models/cuboid)` as the target `x86_64-unknown-linux-musl` does not support these crate types
```

Now the musl binaries, to me at least, are only a nice to have at this point. So @hannobraun we can drop them and only continue to build GNU targets for Linux.

I opened a pull-request: https://github.com/hannobraun/Fornjot/pull/370

hendrikmaus:
There _could_ be a solution though, setting `RUSTFLAGS="-C target-feature=-crt-static"` https://github.com/rust-lang/cargo/issues/8607#issuecomment-676504095

But that resulted in a ton of `cc` link errors for me.

hannobraun:
Thank you for looking into this, @hendrikmaus!

> Now the musl binaries, to me at least, are only a nice to have at this point. So @hannobraun we can drop them and only continue to build GNU targets for Linux.

I agree that they're just a nice-to-have. Dropping them for now is fine, and I think there are two alleviating factors:
1. There are other ways to get binaries that run on most Linux systems, by building against an old glibc, I think. So once that becomes a priority, we'll have options.
2. Building as a dynamic library isn't a long-term solution anyway (#71), so we might be able to re-introduce musl builds later.


[...]

PR #370, opened 2022-03-17
fix(cd): avoid building musl

hendrikmaus:
- one cannot build crate-type cdylib for musl targets
- this also removes `cross` from the show

See https://github.com/hannobraun/Fornjot/actions/runs/1993939647 and https://github.com/hannobraun/Fornjot/issues/369

---

closes https://github.com/hannobraun/Fornjot/issues/369

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -17,20 +17,10 @@ jobs:
     strategy:
       matrix:
         include:
-          # Supported `cross` targets:
-          #   https://github.com/rust-embedded/cross#supported-targets
-
-          # Linux targets
-          - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest, cross: false }
-          - { target: x86_64-unknown-linux-musl, os: ubuntu-latest, cross: true }
-          - { target: aarch64-unknown-linux-musl, os: ubuntu-latest, cross: true }
-
-          # macOS targets
-          - { target: x86_64-apple-darwin, os: macOS-latest, cross: false }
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -17,20 +17,10 @@ jobs:
     strategy:
       matrix:
         include:
-          # Supported `cross` targets:
-          #   https://github.com/rust-embedded/cross#supported-targets
-
-          # Linux targets
-          - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest, cross: false }
-          - { target: x86_64-unknown-linux-musl, os: ubuntu-latest, cross: true }
-          - { target: aarch64-unknown-linux-musl, os: ubuntu-latest, cross: true }
```
hannobraun, line 26: Is dropping support for `aarch64-unknown-linux` necessary? I.e. can it not be built with glibc?

I think it's fine (also see https://github.com/hannobraun/Fornjot/issues/369#issuecomment-1070784881), just asking for clarification.

hannobraun (approved):
Thank you, @hendrikmaus!

Left one question, but will merge immediately.

.github/workflows/cd.yml
hendrikmaus: I missed to check that; there should be a GNU target for ARM on Linux though.

hannobraun: I think it's probably worth supporting, if it's easy. If not, I don't think it's worth spending the effort right now. I've opened #371 to track that.

Thanks!

Issue #141, opened 2022-02-07
Split Fornjot into self-contained crates


hannobraun:
I want Fornjot to one day be the foundation of a thriving ecosystem of related software. To achieve that, it must be composed of self-contained components that can be reused in different contexts.

Right now, Fornjot consists of two components: A host application that loads models, and a library that models use to define geometry. Most of what is relevant lives in the monolithic host application, but I've had modularity in mind from early on. Splitting out the most important components, like the [`kernel`](https://github.com/hannobraun/Fornjot/tree/main/src/kernel) and [`graphics`](https://github.com/hannobraun/Fornjot/tree/main/src/graphics) modules, should be possible with little work, as their interfaces to the rest of the system are pretty clearly defined.

Other components will require some more work to untangle ([`main.rs`](https://github.com/hannobraun/Fornjot/blob/main/src/main.rs) is a bit of a mess right now).

While I see this as absolutely critical long-term, I do not intend to work on it any time soon, unless there is real demand. Having more and smaller crates will introduce friction to the development process, and I'd like to avoid that until there is actual need. If you have such a need, please speak up!

hannobraun:
I started working on this yesterday, but forgot to mention it or to assign myself.

hannobraun:
I had a thought: `fj-debug` should become `fj-interop`, which should also include the mesh (what I intended to extract separately into `fj-mesh`).

The purpose of both `fj-debug` and `fj-mesh` would have been to define a clear interface between the kernel and other components that process its results, but don't need to use the kernel directly (like graphics, export, maybe others). Since that is a single purpose, it makes sense to have a single crate for that. In addition, there might be other such small interop pieces in the futures, and I don't think it would make sense to define more small crates for each of those.

hannobraun:
#105 (and some related issues) have been addressed. I'm back to working on this issue.

hannobraun:
I've done a bunch of work since yesterday, further cleaning up crates that have already been extracted (most of that centered on `fj-interop`.

At this point, I think we're in pretty good shape. There are two more areas of functionality I'd like to extract from `fj-app` before calling this done:
1. `fj-export`: Exporting models to external files (currently only 3MF).
2. `fj-viewer`: Input, camera control, and graphics.

Once those are extracted, I'll could declare victory and call this issue done. I *might* look into splitting up `fj-viewer` further, into separate components for input, camera, and graphics, but I'm not sure that's worth the time investment right now.

hannobraun:
[...]

PR #374, opened 2022-03-17
Extract math code into dedicated crate

hannobraun:
This is progress towards #141. I also took the opportunity to clean up the math code a bit, and improve its documentation.


hannobraun:
I have a fix for the CI failure locally, but GitHub gets an Internal Server Error when I try to push it. Guess I have to wait until that has cleared up.

PR #381, opened 2022-03-18
Extract `fj-host` from `fj-app`

hannobraun:
This is the plugin host code for the models. Recycles the `fj-host`
crate name.


hannobraun:
Forgot to mention the issue that this is working towards addressing: #141

PR #383, opened 2022-03-20
Added platform trait and specific platforms implementations plust Hos…

### LINKED ISSUE #36: Hide platform-specific code behind a single API
There is some [platform-specific code in `model.rs`](https://github.com/hannobraun/Fornjot/blob/8a7f7de0b2995ad34994f2b97fd4ce6f9fc27226/src/model.rs#L26-L38). There's also the [watch event handling code in `main.rs`](https://github.com/hannobraun/Fornjot/blob/8a7f7de0b2995ad34994f2b97fd4ce6f9fc27226/src/main.rs#L133-L140). This is fine for now, but following this approach everywhere will get unwieldy.

I think the following concept would be better:
- Create a new module, `platforms`, that contains a `Platform` trait.
- This trait should define a minimal interface, such that all platform-specific code lives in implementations of that trait.
- Otherwise those trait implementations should contain as little code as possible, to prevent duplication.
- Avoid the use of `#[cfg(...)]` as much as practical. All code should be compiled on all platforms, to prevent errors from creeping in.
- Use of `cfg!(...)` is fine, of course.

ozghimire:
Provides common api with platform traits for host platform. This resolves #36 .

@hannobraun let's review this and I will merge once this is ironed out.

### DIFF
--- a/fj-host/src/platform.rs
+++ b/fj-host/src/platform.rs
@@ -1,0 +1,45 @@
+// Represents platform trait
+trait Platform {
+    fn file_name(name: &str) -> String;
+}
+
+// Represents all platforms supported
+
+// Mac OS
+struct Macos;
+// Windows
+struct Windows;
+// Linux
+struct Unix;
+
+impl Platform for Windows {
+    fn file_name(name: &str) -> String {
+        format!("{}.dll", name)
+    }
+}
+impl Platform for Macos {
+    fn file_name(name: &str) -> String {
+        format!("lib{}.dylib", name)
+    }
+}
+impl Platform for Unix {
+    fn file_name(name: &str) -> String {
+        format!("lib{}.so", name)
+    }
+}
+
+// Represents common apis availiable independent of hosts
+pub struct HostPlatform;
+
+impl HostPlatform {
+    pub fn host_file_name(name: &str) -> String {
+        if cfg!(windows) {
+            Windows::file_name(name)
+        } else if cfg!(target_os = "macos") {
+            Macos::file_name(name)
+        } else {
+            //Unix
+            Unix::file_name(name)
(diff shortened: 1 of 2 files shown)


fj-host/src/platform.rs
```diff
@@ -0,0 +1,45 @@
+// Represents platform trait
+trait Platform {
+    fn file_name(name: &str) -> String;
```
hannobraun, line 3: It might be a good idea to give this a more descriptive name. Maybe `model_lib_file_name`?

fj-host/src/platform.rs
```diff
@@ -0,0 +1,45 @@
...
+        if cfg!(windows) {
+            Windows::file_name(name)
+        } else if cfg!(target_os = "macos") {
+            Macos::file_name(name)
+        } else {
+            //Unix
+            Unix::file_name(name)
+        }
+    }
+}
```
hannobraun, line 45: This is fine, as long as `Platform` only has one method, but if there were more, we'd have to repeat this `if` expression in all of them. How about something like this:
``` rust
pub fn platform() -> impl Platform {
    if cfg!(windows) {
        Windows
    } else if cfg!(target_os = "macos") {
        Macos
    } else {
        Unix
    }
}
```

That would require the trait methods in `Platform` to have a `&self` parameter, so they can be called on the return value, but I think this would be fine.

hannobraun (approved):
Thank you, @ozghimire! This looks great, about what I imagined.

I'm going to merge this as-is, as it provides a clear place for adding future platform-specific infrastructure, which I think is a clear improvement.

In addition, I've left some comments with ideas for improvements. Feel free to address those in a follow-up pull request, if you want!

PR #384, opened 2022-03-21
Refactoring as per review

ozghimire:
Refactored as per #383


hannobraun (approved):
Thank you, @ozghimire, looks great!

PR #387, opened 2022-03-22
ignore models target

anwentec:


anwentec:
The models  are in the cargo workspace. I made a mistake.

hannobraun:
Yeah, the models have been added to the workspace recently, in #368.

PR #389, opened 2022-03-22
:construction_worker: Add models to test suite

### LINKED ISSUE #307: Smoke testing
Sometimes bugs can get introduced that cause Fornjot to panic, if one of the models from the repository is run.I try to test all the models regularly, but sometimes a bug slips through.

It would be great, if as part of the regular CI build, all models would get run. Opening a window to show the model won't work on the CI server, so maybe exporting the model (`cargo run -- --model <model> --export <model>.3mf`) would be a good choice. It should exercise the kernel the same as a regular Fornjot run.

This can be done by extending [`workflows/ci.yml`](https://github.com/hannobraun/Fornjot/blob/main/.github/workflows/ci.yml). I think adding the smoke test to the existing `test` job would be appropriate. Checking the correctness of the exported file would not be necessary. This issue is strictly about a smoke test, to filter out obvious errors.

Labeling as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this should be a relatively straight-forward extension to the existing CI configuration.

Related issues:
- https://github.com/hannobraun/Fornjot/issues/51
- https://github.com/hannobraun/Fornjot/issues/54

homersimpsons:
I couldn't find a way to build the steps from the `models/` directory. Another approach is to have a dedicated `#[test]` that would run the models.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,22 @@ jobs:
       - uses: actions-rs/cargo@4ff6ec2846f6e7217c1a9b0b503506665f134c4b
         with:
           command: test
+      - uses: actions-rs/cargo@4ff6ec2846f6e7217c1a9b0b503506665f134c4b
+        with:
+          command: run
+          args: -- --model cuboid --export cuboid.3mf
+      - uses: actions-rs/cargo@4ff6ec2846f6e7217c1a9b0b503506665f134c4b
+        with:
+          command: run
+          args: -- --model group --export group.3mf
+      - uses: actions-rs/cargo@4ff6ec2846f6e7217c1a9b0b503506665f134c4b
+        with:
+          command: run
+          args: -- --model spacer --export spacer.3mf
+      - uses: actions-rs/cargo@4ff6ec2846f6e7217c1a9b0b503506665f134c4b
+        with:
+          command: run
+          args: -- --model star --export star.3mf
 
   clippy:
     name: Clippy Check


.github/workflows/ci.yml
```diff
@@ -60,6 +60,22 @@ jobs:
...
+          command: run
+          args: -- --model group --export group.3mf
+      - uses: actions-rs/cargo@4ff6ec2846f6e7217c1a9b0b503506665f134c4b
+        with:
+          command: run
+          args: -- --model spacer --export spacer.3mf
+      - uses: actions-rs/cargo@4ff6ec2846f6e7217c1a9b0b503506665f134c4b
+        with:
+          command: run
+          args: -- --model star --export star.3mf
```
hannobraun, line 78: Could we maybe automate this by adding a script to the repository that iterates over all models in `models/`, running each? This is fine for now, but executing all models automatically would be more robust going forward, as models are added, removed, or renamed.

Such a script would have to run on all the platforms we run CI builds on. There are many ways to do this. [Nushell](https://www.nushell.sh/) has been suggested for this purpose in the past.

hannobraun (approved):
Thank you, @homersimpsons! This is great!

I've left a comment with a suggestion, but since this is already a big improvement, I'm going to merge it right away. Feel free to address my comment in a follow-up PR, if you like!

I'd like to leave #307 open, until we have some mechanism to execute the models automatically, as per my comment. I'm going to edit the PR description accordingly, before I merge.

PR #390, opened 2022-03-23
Add methods to construct unit vectors

hannobraun:
This came out of some cleanup work that, so far, didn't pan out.


hannobraun:
This has been sitting open since yesterday, since GitHub was having problems yet again, and the CI build refused to run. Looks like it's working now.

PR #391, opened 2022-03-24
Create GitHub Release via CD workflow

hendrikmaus:
This change-set activates the release-operator in the CD workflow and updates the release process document.

Related issues: https://github.com/hannobraun/Fornjot/issues/104

PR #392, opened 2022-03-24
Upload artifacts using a wildcard

hendrikmaus:
This change-set updates the `actions/upload-artifact` Action to `v3`, which supports a wildcard match on the `path` field. This allows us to simplify the workflow.

I renamed some steps for increased user experience while looking at the workflow summary.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -38,14 +38,13 @@ jobs:
       - name: Cache
         uses: Swatinem/rust-cache@v1
 
-      - name: Compile
+      - name: Binaries | Compile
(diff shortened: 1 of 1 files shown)


hannobraun (changes requested):
Thank you, @hendrikmaus!

Looks good to me, although again, I'm not sure how much good that does when I'm reviewing CI/CD stuff :smile: 

Conflicts with #391, which I just merged. Needs a rebase before it can be merged.

hendrikmaus:
You're welcome. Four eyes are always better than two :)

Conflict resolved.

hannobraun (approved):
Thanks again!

PR #393, opened 2022-03-24
Improve caching in CI and CD

hendrikmaus:
This change-set changes the `actions-rs/toolchain@v1` Action temporarily to a fork that supports `rust-toolchain.toml` as intended. The issue on the upstream Action is open, but not merged yet https://github.com/actions-rs/toolchain/pull/209. Once merged and released, we can switch back to the upstream Action in all workflow files.

In the CD workflow, the target is now included in the cache key. In my experiments, that improved caching, especially for the apple targets. Before, the aarch apple target would run with the x86_64 apple cache.

- [x] also adapt the changes to #391

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -38,6 +38,8 @@ jobs:
 
       - name: Cache
         uses: Swatinem/rust-cache@v1
+        with:
+          key: ${{ matrix.target }}
 
       - name: Compile
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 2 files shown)


hannobraun (approved):
Thank you, @hendrikmaus, looks good to me! (Disclaimers about my CI/CD non-expertise apply.)

I did a rebase (#391 was merged in the meantime). The PR applies cleanly, CI runs, and seems to use the right version. Merging!

Issue #395, opened 2022-03-24
Continuous deployment fails


hannobraun:
See https://github.com/hannobraun/Fornjot/actions/runs/2033227732.

There seems to be a syntax error in the config:
```
ParserError: D:\a\_temp\bbb682d5-c527-4e38-a45a-773e612d3cef.ps1:7
Line |
   7 |  if [[ "${RUNNER_OS}" == "Windows" ]]; then
     |    ~
     | Missing '(' after 'if' in if statement.
```

cc @hendrikmaus

hendrikmaus:
I suppose it tries to use Powershell 🙈 I opened a pull-request to let all `run` steps always use the `bash` shell.

hendrikmaus:
Works 🥳

PR #396, opened 2022-03-24
Set the default shell to bash

hendrikmaus:
Closes https://github.com/hannobraun/Fornjot/issues/395


hannobraun (approved):
Thank you, @hendrikmaus! Let's try it!

Issue #157, opened 2022-02-09
Fix wgpu warnings


hannobraun:
Since I re-enabled logging recently, I've been noticing warnings being logged by wgpu. Here's the output from a typical run:

```
2022-02-09T13:27:48.037289Z  WARN wgpu_hal::vulkan::instance: Unable to find layer: VK_LAYER_KHRONOS_validation    
2022-02-09T13:27:48.039542Z  INFO wgpu_hal::vulkan::instance: Instance version: 0x4020cb    
2022-02-09T13:27:48.039672Z  INFO wgpu_hal::vulkan::instance: Enabling debug utils    
2022-02-09T13:27:48.039793Z  INFO wgpu_hal::vulkan::instance: Enabling device properties2    
MESA-INTEL: warning: Haswell Vulkan support is incomplete
2022-02-09T13:27:48.045609Z  INFO wgpu_core::instance: Adapter Vulkan AdapterInfo { name: "Intel(R) HD Graphics 4600 (HSW GT2)", vendor: 32902, device: 1042, device_type: IntegratedGpu, backend: Vulkan }    
2022-02-09T13:27:48.076110Z  INFO wgpu_hal::vulkan::adapter: Private capabilities: PrivateCapabilities { flip_y_requires_shift: true, imageless_framebuffers: true, image_view_usage: true, timeline_semaphores: true, texture_d24: true, texture_d24_s8: true, can_present: true, non_coherent_map_mask: 63, robust_buffer_access: true, robust_image_access: true }    
2022-02-09T13:27:48.079807Z  INFO wgpu_core::device: configuring surface with SurfaceConfiguration { usage: RENDER_ATTACHMENT, format: Bgra8UnormSrgb, width: 800, height: 600, present_mode: Fifo }    
2022-02-09T13:27:48.082956Z  INFO wgpu_core::device: Created texture Valid((0, 1, Vulkan)) with TextureDescriptor { label: None, size: Extent3d { width: 800, height: 600, depth_or_array_layers: 1 }, mip_level_count: 1, sample_count: 1, dimension: D2, format: Depth32Float, usage: RENDER_ATTACHMENT }    
2022-02-09T13:27:48.090101Z  INFO wgpu_core::device: Created buffer Valid((0, 1, Vulkan)) with BufferDescriptor { label: None, size: 128, usage: COPY_DST | UNIFORM, mapped_at_creation: true }    
[...]

PR #397, opened 2022-03-24
fix(config_ui): fix warning because of missing finish() call

liubog2008:
Fix warning as below.

```
  2022-03-23T08:06:42.578774Z  WARN wgpu_core::device::queue: Dropped buffer has a pending mapping.
```

See example in https://github.com/hecrj/wgpu_glyph/blob/master/examples/hello.rs

Related issue: https://github.com/hannobraun/Fornjot/issues/157


hannobraun (approved):
Thank you, @liubog2008, this is great!

I didn't consider that those warnings could be coming from wrong use of `wgpu_glyph`!

PR #408, opened 2022-03-29
Add sub-command support to the release-operator

hendrikmaus:
This change-set moves the existing functionality of the release-operator into a sub-command. See https://github.com/hannobraun/Fornjot/issues/104#issuecomment-1079002976 for the relevant discussion.

PR #409, opened 2022-03-29
fix: align output names

hendrikmaus:
When no release was detected, the output name was incorrect due to the
usage of a stringly typed identifier. I implemented the outputs as an
enum to overcome the issue.

### DIFF
--- a/release-operator/src/github.rs
+++ b/release-operator/src/github.rs
@@ -1,5 +1,6 @@
 use cmd_lib::run_fun;
 use serde::Deserialize;
+use crate::release::Outputs;
 
 #[derive(Deserialize, Debug, Clone)]
 pub struct PullRequest {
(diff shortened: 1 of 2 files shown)


hendrikmaus:
With this fix in place, only the YAML file in GitHub Actions is still stringly typed, i.e. https://github.com/hannobraun/Fornjot/blob/main/.github/workflows/cd.yml#L110

I do not yet have a good idea how to overcome that.

hendrikmaus:
For easier glance, this was the bug:
image

hannobraun (approved):
Thank you @hendrikmaus, looks good!

> With this fix in place, only the YAML file in GitHub Actions is still stringly typed, i.e. https://github.com/hannobraun/Fornjot/blob/main/.github/workflows/cd.yml#L110
>
> I do not yet have a good idea how to overcome that.

Maybe nothing we an do about it, except maybe move more CD logic out of GitHub Actions and into the release operator. No idea, if this is practical or even desirable.

hannobraun:
By the way, I had to rebase due to #408 being merged. Will merge once CI is green.

PR #437, opened 2022-04-07
Fix issue on my new AMD GPU

hannobraun:
I got my new PC today, which has an AMD CPU/GPU. The old one had Intel, as does my laptop.

This pull request fixes a crash with that hardware, by logging a warning instead of panicking on a draw error. It then fixes the large number of warnings printed by switching the present mode from `PresentMode::Fifo` to `PresentMode::Mailbox`.


hannobraun:
Also works fine on my laptop. Merging.

PR #463, opened 2022-04-12
Move parameter parsing out of `main`

hannobraun:


hannobraun:
Man, looks like I totally missed that this commit doesn't work. I thought I tested it, but obviously not.

I found this lying around in a local development branch that I hadn't touched in a while, and figured, looks good, let's get it in. In the meantime, I've found some more stuff that I need to clean up around `main` to make progress on #141, so I'm just going to close this, and might re-submit it (hopefully fixed, by then) in a future PR together with more cleanup of `main`.

Issue #104, opened 2022-01-28
Automate release procedure


hannobraun:
It would be great to largely automate the [release procedure](https://github.com/hannobraun/Fornjot/blob/main/RELEASES.md).

There are basically just three parts that need to be manual:
1. Choose the version number.
2. Update the changelog.
3. Write a release announcement.

It would be great, if somehow we could provide these ingredients in a commit (maybe by adding an entry to a `releases.toml` file that specifies the version number and links the release announcement), then have an automated action be triggered by that, that takes care of everything else.

I'm not sure how practical this is. For example, I don't know if such a trigger can be specified. But it would be great to have.

Labeling as https://github.com/hannobraun/Fornjot/labels/help%20wanted, as I'm no expert in GitHub Actions. If someone who knows more happens to drop by and could leave some insights about the feasibility of this, that would be very appreciated!

hannobraun:
@jamwaffles suggested [`cargo-release`](https://crates.io/crates/cargo-release). I haven't looked into it, but it's certainly worth checking out.

hendrikmaus:
Yes, let's tackle #57 first. We should be able to store the binaries on each workflow run, using the artifacts feature. That way, no release is needed to host them.

hannobraun:
@jamwaffles made me aware of release-plz: https://github.com/MarcoIeni/release-plz

I haven't had the chance yet to check it out, but it seems worth looking into.

hendrikmaus:
Nice, I was researching in the same direction in the last days. I adopted the "original" release please in this crate https://github.com/hendrikmaus/kube-leader-election but, at a glance, the new proposal is even better.

Are you also ok with adopting conventional commits? I am trying right now in every project and so far it is beneficial.

hannobraun:
> Are you also ok with adopting conventional commits? I am trying right now in every project and so far it is beneficial.

For the following comment, I'm assuming we're talking about [this specification](https://www.conventionalcommits.org/). I wasn't aware that existed, but I tried a similar technique a few years back, to generate changelogs automatically. Based on that experience, and a short reading of the specification, I'm against enforcing it (although I have no problem merging commits that follow that format).

The specification has a section titled [Why Use Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.2/#why-use-conventional-commits). I'm going to go through each of the items to explain my position.

---

> Automatically generating CHANGELOGs.

[...]

PR #469, opened 2022-04-13
Determine if the current crate is ahead or behind the registry

hendrikmaus:
This change-set let's `release-operator` cover the case in which the crates.io registry has a newer version than ours in scope.

Related issue: https://github.com/hannobraun/Fornjot/issues/104

### DIFF
--- a/release-operator/Cargo.toml
+++ b/release-operator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "release-operator"
-version = "0.2.0"
+version = "0.3.0"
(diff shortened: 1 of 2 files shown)


hendrikmaus:
I will crate a second change-set, based on this branch, that uses the crates.io API to gather the crate information. That way, we get to see all versions and can determine if the one in scope hasn't been published at all.

hannobraun (approved):
Thank you, @hendrikmaus, looks great!

PR #470, opened 2022-04-13
Switch to the crates.io HTTP API

hendrikmaus:
This change-set uses the crates.io HTTP API instead of a shell-out to `cargo search` for looking up the crate in scope in the upstream registry.

FYI: In a next change-set, the logic will be adapted to check the entire list of versions in the registry. This will be done if the operator encounters that the current local crate version is behind the registry - i.e. the registry has a newer version. Then it can iterate through all versions that the upstream registry has to determine if the current local version should be published anyway.

### DIFF
--- a/release-operator/src/registry.rs
+++ b/release-operator/src/registry.rs
@@ -77,7 +77,8 @@ impl Crate {
                 .context("search crates.io for published crate version")?;
             log::debug!("{self} found as {version} on their side");
 
-            semver::Version::from_str(&version).context("parse their version")?
+            semver::Version::from_str(&version)
+                .context("parse their version")?
         };
 
         let ours = {
@@ -111,7 +112,7 @@ impl Crate {
 
         if ours < theirs {
             log::warn!("{self} has already been published as {ours}, which is a newer version");
-            return Ok(CrateState::Behind)
+            return Ok(CrateState::Behind);
         }
 
         Ok(CrateState::Ahead)
--- a/release-operator/src/registry.rs
+++ b/release-operator/src/registry.rs
@@ -1,5 +1,6 @@
 use anyhow::{anyhow, Context};
 use secstr::SecStr;
+use serde::Deserialize;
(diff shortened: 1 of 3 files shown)


release-operator/src/registry.rs
```diff
@@ -64,21 +65,51 @@ impl Crate {
...
+                ));
+            }
+
+            let versions =
+                serde_json::from_str::<CrateVersions>(resp.text()?.as_str())
+                    .context("deserializing crates.io response")?;
 
-            semver::Version::from_str(&version)
-                .context("parse their version")?
+            versions.versions.get(0).unwrap().version.to_owned()
```
hendrikmaus, line 112: Todo for me: get rid of this `unwrap`

hannobraun: `.ok_or_else(|| anyhow!("error message"))` should do it?

hannobraun (approved):
Thank you, @hendrikmaus, looks good to me!

One note: I generally prefer rebases to merges, unless more changes are made after a pull request has already been reviewed. Often it doesn't matter, but this was one case where there was a slight mental overhead, because I had to identify if those were commits I had already seen before, or if there were any changes.

Not a big deal; just saying that in cases like this, a rebase would have made the review process (very) slightly easier.

> FYI: In a next change-set, the logic will be adapted to check the entire list of versions in the registry. This will be done if the operator encounters that the current local crate version is behind the registry - i.e. the registry has a newer version. Then it can iterate through all versions that the upstream registry has to determine if the current local version should be published anyway.

That would definitely be useful, if we ever decide to maintain release branches, but it also gets my YAGNI sense tingling :grin: 

I'm certainly not opposed to it, but I also wouldn't mind to leave this case unaddressed for now.

release-operator/src/registry.rs
hendrikmaus: Exactly what I had in mind. No idea why I didn't do that right away 🤷🏻‍♂️

Issue #54, opened 2022-01-20
Add integration test for triangle mesh quality


hannobraun:
To prevent any regressions to the quality of the generated triangle meshes, it would be great to have some kind of test suite that builds a triangle mesh for each of the models in this repository, and checks that mesh for quality. Either by using parts of Fornjot as a library and checking internal data structures directly, or by exporting a 3MF/STL/... file and checking that.

This test suite wouldn't check if the generated mesh was correct for the specific model, but would just check the mesh itself for correctness:
- Is the mesh manifold?
- Are triangle normals consistent?
- Are there duplicate or overlapping triangles?

And whatever other ways a mesh can be wrong (please comment, if you can think of another one).

Maybe there are libraries in the Rust ecosystem we could use for this, or maybe there's some suitable command-line tool. (I do know that slicers routinely do this kind of check, so maybe it's possible to use one from the command-line for this purpose. But a Rust library would be nicer, all else being equal.)

hannobraun:
I've started looking into this, as part of addressing #339 and #451. Most likely, I'll address this by including a 3MF validator in the CI build somehow.

PR #475, opened 2022-04-13
Add `crates/` and `tools/` directories

hannobraun:
See commit messages for justification. The reason I'm doing this now, is that I'm working on adding a second tool crate, as part of #54.

What do you think, @hendrikmaus? Since this affects release-operator and possibly the further automation of the relase process, both topics you've worked on exclusively, I'd love to get your input.

### DIFF
--- a/crates/fj-app/src/main.rs
+++ b/crates/fj-app/src/main.rs
@@ -3,7 +3,7 @@ mod config;
 
 use std::path::PathBuf;
 
-use anyhow::anyhow;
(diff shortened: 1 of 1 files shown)


hendrikmaus:
From here it looks all fine and the structure makes sense to me. You've changed the relevant spots where paths needed adjustment already 👍🏻

hannobraun:
Thanks for the feedback! I'm currently figuring out what I need to change to make the "file not found" error go away, then this is ready for merging.

Issue #451, opened 2022-04-11
3MF export is partially broken because of triangle orientation


hannobraun:
Here's a 3D-printed version of the star model:
![Picture of a black, star-shaped 3D print](https://braun-odw.ems.host/_matrix/media/r0/download/matrix.org/HrsEydrTjKXPSwzdGXtiQzZM)

The star-shaped hole is missing. This is most likely related to validation errors of the exported 3MF file:

![Screenshot from a 3MF validation tool, showing a "bad triangle orientation" error](https://braun-odw.ems.host/_matrix/media/r0/download/matrix.org/HbAmNSKIrxTavLlezsyludnT)

This is not surprising, as the kernel doesn't really have any concept of surface orientation. What wasn't clear so far, is that this affects the validity of exported 3MF files.

#339 is related, as both bugs have the same root cause. Another related (and now closed) issue is #173.

We should prevent such issues outright, by adding a 3MF validation step to the CI build (see #54).

Thanks to [@dcompoze](https://github.com/dcompoze) for reporting this on [Matrix](https://matrix.to/#/#fornjot:braun-odw.eu)!

hannobraun:
An additional thought: As part of addressing this, it would probably be a good idea to revert #289, to make orientation bugs more visible.

hannobraun:
I've started looking into this.

PR #481, opened 2022-04-14
Expand `export-validator` to actually validate exported files

### LINKED ISSUE #54: Add integration test for triangle mesh quality

hannobraun:
This pull request can't be merged yet, because this validation fails (see #451). I'm going to address that next, and once I have, `export-validator` will make sure there are not going to be regressions.

Close #54


hannobraun:
I've extracted the 3MF validation into its own job, which means none of the required jobs are failing any more, which means this can actually be merged now!

Issue #478, opened 2022-04-14
CI build should fail, if crates are missing `![deny(missing_docs]`


hannobraun:
As of #475 crates are organized into different directories. All crates that are targeted at ecosystem or end users are in [`crates/`](https://github.com/hannobraun/Fornjot/tree/main/crates), internal tools are in [`tools/`](https://github.com/hannobraun/Fornjot/tree/main/tools), and as before, all models are still in [`models/`](https://github.com/hannobraun/Fornjot/tree/main/models).

All user-facing crates should be fully documented. To prevent mistakes, the CI build should check that all crates in the `crates/` directory have a `![deny(missing_docs)]` in their `lib.rs`.

This could be added to the [CI configuration](https://github.com/hannobraun/Fornjot/blob/main/.github/workflows/ci.yml) as an additional job. That job can run only on Linux, like the *Format* and *Clippy Check* jobs. If this isn't easy to do only with GitHub Actions, we can add a new Rust program to `tools/`, and call that from the job using `actions-rs/cargo` (other jobs are already doing this).

Labeling as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, since this doesn't require lots of knowledge of Fornjot, and could serve as an introduction to the project.

Related issue: #477

freylint:
I'm someone who tinkered with porting parts of openscad to rust and am currently considering picking this up as a first issue.

The requirements only force full documentation of user facing crates time of PR. This gives us some ambiguity in how it can be implemented. As I'm new to this project, I'd love for some developer insight into my thoughts on my solutions to this issue.

#### A lint only mechanism
We could try a pass of cargo clippy with all lints but `missing_docs` disabled. This would cause the build to fail on CI without failing on the developer machine.

I've seen a lot of contention on denying this lint, as it can make hacking on the codebase frustrating for some.

#### Parsing Cargo.toml and crawling the crates [^1]
We could alse parse the cargo.toml and check for the lint until some set of code that would forbid top level lint declarations. We could then force every crate in the repo to apply the `missing_docs` lint for every stage of development and CI.

#### A Compromise [^2]
I suggest using the _parse and crawl_ approach to instead enforce a warn of `missing_docs` to avoid workflow disruption and use the _lint only_ mechanism at CI time to fail any PRs which haven't followed the warning.

[^1]: a crawler like this should be integrated with related CI improvements in #477
[^2]: #269 suggests that this is an approach which the maintainers don't disagree with.

[...]

PR #486, opened 2022-04-18
Fail PRs which fail the missing_docs lint.

freylint:
Partially addresses #478.
Causes PRs which fail the `missing_docs` lint to fail CI regardless of if the lint is enabled in the crate. 

This could be adapted to bring the project in line with the linting suggestions of the [Rust Unnoficial Patterns Repo](https://github.com/rust-unofficial/patterns/blob/main/anti_patterns/deny-warnings.md#alternatives) as well as denying all warnings.

I'm leaning towards not denying all warnings as github CI seems to support the display of warnings as separate from errors. This would make warnings appear as errors on review.
I suggest investigating CI settings to forbid merging of PRs which produce warnings.

_NOTE_: A number of existing components fail this lint, causing the CI pipeline to fail. This needs to be addressed before merge.


hannobraun (commented):
Thank you for the pull request, @freylint! I'll follow [your recommendation](https://github.com/hannobraun/Fornjot/pull/487) of not merging this yet. I'm going to convert the pull request into a draft, to clarify the status meanwhile.

Other than that, I'm happy with what I'm seeing here! Happy to merge once CI is green.

> I'm leaning towards not denying all warnings as github CI seems to support the display of warnings as separate from errors. This change would make warnings appear as errors on review.

My experience has been that these warning annotations can be easy to ignore, both by contributors and by overburdened maintainers. I'm pretty much on top of things here (it's kind of my job, after all :grin:), so maybe using annotations and not failing the build would be fine in this case.

I'm still leaning towards denying all warnings, as that sends a clearer message, but that's not a strong preference, as of now.

> I suggest investigating CI settings to forbid merging of PRs which produce warnings as an alternative.

I've taken a look at the settings and can't find anything, although I'm not confident I'm looking in the right places. I don't see how that would be any different from just failing the CI build though, in practical terms.

hannobraun:
Rebased, now that #487 is merged. Clippy is still failing, but it doesn't seem documentation-related, but rather related to a change being made to the Clippy action.

Not sure I understand what is going on, but I'll just risk merging this. If that turns out to be a mistake, we can deal with that later.

hannobraun:
Oh, wait, I can't merge this without changing the branch protection config, which I don't want to do, because restoring it to its current state is going to be a pain.

I'll dig into the failure later. There might be some way to get this merged without messing up the branch protection configuration.

hannobraun:
The CI error message points to this issue: https://github.com/actions-rs/clippy-check/issues/2

If I understand it correctly, `clippy-check` should fall back to working behavior, if run from a fork. So I don't understand why it won't work.

I will look into this further once I'm back from vacation (my motivation has limits :smile:). If someone else wants to look into this in the meantime, that would be very welcome.

hannobraun:
Just rebased this pull request to check whether the error is still there, and it is.

In my previous comment, I wondered why `clippy-check` isn't falling back to working behavior. And indeed, other pull requests don't seem to have any trouble with displaying Clippy errors/warnings, despite showing the same errors and warnings: https://github.com/hannobraun/Fornjot/runs/6257824051?check_suite_focus=true#step:6:329

I now think that the failure probably has nothing to do with `clippy-check`. When running `cargo clippy --all-features -- -D missing_docs` locally, it complains about missing documentation in `fj-app`. I think it's likely that this is the cause of the failure here too. Why `clippy-check` doesn't display this error message, I don't know, and I'm not inclined right now to hunt down CI mysteries.

A more relevant insight I gained from this is, that the approach chosen in this pull request is probably wrong, in my opinion. That `fj-app` has no API documentation is irrelevant, because it has no API. It's an application. We could add doc comments to satisfy that warning, or we could do something I [suggested before](https://github.com/hannobraun/Fornjot/issues/478#issuecomment-1100832345): Use `#![warn(missing_docs)]` in all crates where that makes sense, and run the CI build with `-D warnings`

What do you think, @freylint?

PR #493, opened 2022-04-22
Convert `Arc` back into `Circle`; improve circle representation

hannobraun:
This is an overall improvement to circle representation, fixing a bunch of issues, some conceptual, some real bugs. See commit messages for more information.

We also lose the capability to represent arcs. This capability has only been there for a very short time (since #483) and wasn't used anywhere. It can be re-added later.


hannobraun:
The model I added to test the new circle transform capabilities results in an invalid 3MF file, when exported. Since the model would have been even more wrong before, meaning the changes here are strictly an improvement, I'm going to remove the commit that adds the model, merge this pull request, then open a new issue.

For the record, here's a screenshot of the new model without the improvements from this PR:
Screenshot from 2022-04-22 13-02-20

And the same model with the improvements:
transform

hannobraun:
Opened #494.

PR #497, opened 2022-04-24
Relax interface of `fj-viewer::{window, input::Handler}`

freylint:
Relaxes the internal interface of `fj-viewer` making it easier to implement alternative input/windowing backends for platforms like web/mobile or for testing / CI.

### DIFF
--- a/crates/fj-viewer/src/input/handler.rs
+++ b/crates/fj-viewer/src/input/handler.rs
@@ -101,7 +101,7 @@ impl Handler<Winit> {
         self.cursor = Some(cursor);
     }
 
-    /// Updates `state` and `focus_point` when mouse is clickled.
+    /// Updates `state` and `focus_point` when mouse is clicked.
     pub fn handle_mouse_input(
         &mut self,
         button: MouseButton,
--- a/crates/fj-viewer/src/input/handler.rs
+++ b/crates/fj-viewer/src/input/handler.rs
@@ -21,6 +21,8 @@ use super::{movement::Movement, rotation::Rotation, zoom::Zoom};
 /// Input handling abstraction
 ///
 /// Takes user input and applies them to application state.
+///
+#[doc = include_str!("../../docs/phantom_type.md")]
 pub struct Handler<T> {
     cursor: Option<PhysicalPosition<f64>>,
 
--- a/crates/fj-viewer/src/input/handler.rs
+++ b/crates/fj-viewer/src/input/handler.rs
@@ -81,7 +81,7 @@ impl Handler {
         &mut self,
         cursor: PhysicalPosition<f64>,
         camera: &mut Camera,
-        window: &Window,
+        window: &Window<winit::window::Window>,
     ) {
         if let Some(previous) = self.cursor {
             let diff_x = cursor.x - previous.x;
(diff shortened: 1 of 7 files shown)


crates/fj-viewer/src/input/handler.rs
```diff
@@ -13,22 +13,27 @@ use winit::{
...
 };
 
 use super::{movement::Movement, rotation::Rotation, zoom::Zoom};
 
 /// Input handling abstraction
 ///
 /// Takes user input and applies them to application state.
-pub struct Handler {
+///
+#[doc = include_str!("../../docs/phantom_type.md")]
```
hannobraun, line 25: That's a neat trick that I didn't know about. Nice!

hannobraun (commented):
Thank you for the pull request, @freylint!

I understand what this pull request does, but I don't understand the justification. You mention alternative backends for web/mobile, but winit supports these targets. I'm not aware of any reasons, why winit wouldn't be sufficient for our future needs.

You also mention testing/CI. Do you have anything specific in mind?

This pull request, by itself, seems like a needless (if small) increase in complexity. This is not a problem. I merge pull requests like this all the time, while I'm working towards a larger goal. I just want to make sure I understand what this is working towards.

freylint:
Apologies as personal life got in the way of this. I was investigating making a web based frontend for fornjot that incorporated a built in editor.

This change was made to allow mock implementations, increase expressiveness, and investigate the usage of wgpu directly.

I think it's worth clarifying here that this change only increases the complexity of the type signature and the constructor while enabling backend genericism and specialization and test mocking.
I don't think it's my place to make this decision. If you want to insist that all UI be done through winit that is your decision, I just caution against that as I suspect a custom GPU accelerated UI would be quite burdensome and prone to being hard to find contributors for. This is also duplicate effort with several projects which specialize in this exact problem, eg: GTK, WinApi, Wry.

hannobraun:
Thank you for your reply, @freylint!

> If you want to insist that all UI be done through winit that is your decision, I just caution against that as I suspect a custom GPU accelerated UI would be quite burdensome and prone to being hard to find contributors for. This is also duplicate effort with several projects which specialize in this exact problem, eg: GTK, WinApi, Wry.
> Of course this is a whole separate discussion.

I'm not insisting that we keep using winit. But I was under the impression that any alternative solution would either replace everything outright (like Bevy, to use that example again, which would be a complete solution with its own input/windowing/graphics stack), or would provide no reason to replace winit without replacing the input handling (like a higher-level graphics engine to replace wgpu).

Your mention of the kind of alternatives you're thinking about (GTK, WinAPI, Wry; the last of which I didn't know about) clears up the purpose of this pull request for me.

Although I'd like to add, I don't think any of these could be a replacement for winit *in this context*. `fj-viewer` should be able to run in a browser eventually. I don't think any of those alternatives can support that, but winit can.

Now, it's totally reasonable to want to embed Fornjot in a GTK application, for example. If winit conflicts with that, then maybe that means that `fj-viewer` should be trimmed down, moving the windowing aspect out of it. I think that makes a lot of sense. I'll open an issue with more thoughts later.

---

This still leaves me with the question of whether to merge this change. Yes, it increases generality, but it's also kind of trivial and doesn't provide anything that you couldn't build in 10 minutes, right before you actually need it.

As I said before, I merge changes like this all the time. Something like "I'm done for the day, and wanted to get this merged to keep the diff with my local branch as small as reasonable" is a perfectly fine reason to get something like this merged. But this is also the kind of change that I later remove without hesitation, if it turns out it wasn't needed after all.

So again, if this is helping you with something you're currently working on, great, happy to merge. I don't want to merge it, if it's just something that *could* be useful. [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), you know?

PR #498, opened 2022-04-24
Add vstudio and vscode ide configs to .gitignore.

freylint:
This makes it nicer to work on the project for people who don't use IntelliJ IDEs.
Also add a nested dir wildcard before the ignore to also ignore when these are created in individual crates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,4 @@
 
 # IDE Configurations
 **/.vs/
-**/.vscode/
 **/.idea/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,5 +1,7 @@
 # Cargo
 /target
 
-# IntelliJ IDEA
-/.idea
+# IDE Configurations
+**/.vs/
+**/.vscode/
```
hannobraun, line 6: There already is VS Code configuration in the repository, which I'm actively using.
```suggestion
```

hannobraun (approved):

Issue #518, opened 2022-05-03
Clean up use of math code


hannobraun:
Fornjot was originally built on [nalgebra](https://nalgebra.org/) and [Parry](https://www.parry.rs/). It still is, to a large extent, but we have grown our own math types which live in the `fj-math` crate. Those math types still use nalgebra/Parry under the hood, but their API is tailored towards Fornjot's needs, which reduces complexity and increases reliability overall.

While the kernel and other critical components are largely ported to `fj-math`, `fj-viewer` (and possibly other crates) still make heavy use of nalgrabra/Parry directly. It would be nicer if all Fornjot crates used `fj-math` for their math needs, as this would reduce the number of dependencies and overall complexity.

Specifically, the following work should be done, as part of this issue:
1. Replace all uses of nalgebra and Parry outside of `fj-math` with `fj-math`. Extend `fj-math` as required.
2. Remove nalgebra and Parry dependencies from all crates, except `fj-math`.
3. Simplify the `fj-math` API by removing all nalgebra/Parry interop features, making their use an implementation detail.

I think this issue could serve as a nice introduction to Fornjot, as it involves working on multiple Fornjot crates, without requiring detailed knowledge of any of them. Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue.

hannobraun:
Thank you, @connor-lennox, sounds great! Let me know, if there's anything I can help with.

One additional note: If I remember correctly, Parry is used in `fj-viewer` for computing the focus point, i.e. the point on the model that the mouse cursor is currently hovering over. This is used for input processing / camera control.

I'm currently considering simplifying the input code. See #522 for full context. So if any of the Parry-using code in `fj-viewer` is causing you any trouble, just be aware that removing/simplifying it is an option.

hannobraun:
Addressed in #523.

PR #523, opened 2022-05-05
Math cleanup

connor-lennox:
Addresses #518 .

The work here is split into two main sections: 

1. Additions to `fj-math` that forward behaviour from `nalgebra` and `parry`
2. Reworking components in `fj-viewer` and `fj-kernel` to use only the types from `fj-math`

One major change is the conversion of `fj-math::Transform` from wrapping a `TGeneral` to `TAffine` `nalgebra::Transform`. This doesn't have any impact on anything yet (since the `TGeneral` capabilities were not being used anywhere) but it's something to consider for if `fj-math::Transform` becomes more generalized in the future.

As always, happy to discuss these changes either here or on Matrix. This is (one of) my first forays into open-source Rust so please don't be afraid to pick it apart and show me how things could be better implemented.


hannobraun:
Thank you for the pull request, @connor-lennox! I've had a quick look, and this looks good. I'll do a proper review later today.

> This is (one of) my first forays into open-source Rust so please don't be afraid to pick it apart and show me how things could be better implemented.

Understood. I'll be extra nitpicky then :grin:

crates/fj-math/src/point.rs
```diff
@@ -58,6 +58,11 @@ impl<const D: usize> Point<D> {
             coords: self.coords.to_xyz(),
         }
     }
+
+    /// Gives the distance between two Points.
```
hannobraun, line 62: Nitpicky comment: I'd either use "point" as a normal word with proper capitalization, or make it clear I'm referring to the type `` `Point` ``, by adding backticks. And if we weren't already looking at the documentation of `Point`, I'd make it a link, ``[`Point`]``. If `Point` is in scope, rustdoc will make that link work automatically in the generated API reference).

crates/fj-viewer/src/graphics/transform.rs
```diff
@@ -1,15 +1,15 @@
 use bytemuck::{Pod, Zeroable};
-use nalgebra::{Matrix4, Perspective3};
 
 use crate::camera::Camera;
+use fj_math::Transform as fjTransform;
```
hannobraun, line 4: Maybe just remove this instead, and use it as `fj_math::Transform`? That's more verbose, but also more clear, I think.

crates/fj-viewer/src/graphics/vertices.rs
```diff
@@ -47,19 +51,41 @@ impl Vertices {
...
+                    position.x.into_f64(),
+                    position.y.into_f64() - d,
+                    position.z.into_f64(),
+                ]),
+                Point::from_array([
+                    position.x.into_f64(),
+                    position.y.into_f64() + d,
+                    position.z.into_f64(),
+                ]),
+            ],
```
hannobraun, line 88: See comment above:
```suggestion
            [
                position - Vector::from([0., d, 0.]),
                position + Vector::from([0., d., 0.]),
                ]),
            ],
```

crates/fj-viewer/src/graphics/vertices.rs
```diff
@@ -29,12 +29,16 @@ impl Vertices {
...
         normal: [f32; 3],
         color: [f32; 4],
     ) {
         let line = line.into_iter().map(|point| Vertex {
-            position: [point.x as f32, point.y as f32, point.z as f32],
+            position: [
+                point.x.into_f32(),
+                point.y.into_f32(),
+                point.z.into_f32(),
+            ],
```
hannobraun, line 41: Can be shortened:
```suggestion
            position: point.coords.components.map(|scalar| scalar.into_f32()),
```

crates/fj-viewer/src/input/handler.rs
```diff
@@ -146,7 +146,12 @@ impl Handler {
         self.zoom.discard_old_events(now);
         self.zoom.update_speed(now, delta_t, focus_point, camera);
 
-        camera.translation.z -= self.zoom.speed();
+        camera.translation = camera.translation
+            * Transform::translation(Vector::from_components_f64([
+                0.0,
+                0.0,
+                self.zoom.speed(),
+            ]));
```
hannobraun, line 154: This inverts the mouse wheel behavior. Previously, the metaphor was "mouse wheel moves model", this switches it to "mouse wheel moves camera". Also, the `Vector` construction can be shortened to `Vector::from`

This should do it, I think:
```suggestion
        camera.translation = camera.translation
            * Transform::translation(Vector::from([
                0.0,
                0.0,
                -self.zoom.speed(),
            ]));
```

crates/fj-viewer/src/input/movement.rs
```diff
@@ -43,14 +43,17 @@ impl Movement {
...
                 let diff = (cursor - previous) * d2 / d1;
                 let offset = camera.camera_to_model().transform_vector(&diff);
 
-                camera.translation.x += offset.x;
-                camera.translation.y += offset.y;
+                camera.translation =
+                    camera.translation
+                        * Transform::translation(Vector::from_components_f64(
+                            [offset.x.into(), offset.y.into(), 0.0],
+                        ));
```
hannobraun, line 56: Can be shortened, I think:
```suggestion
                        * Transform::translation(Vector::from(
                            [offset.x, offset.y, 0.0],
                        ));
```

Or even shorter, although maybe less clear:
```suggestion
                        * Transform::translation(offset.to_xyz());
```

crates/fj-viewer/src/input/rotation.rs
```diff
@@ -26,21 +26,24 @@ impl Rotation {
 
     pub fn apply(&self, diff_x: f64, diff_y: f64, camera: &mut Camera) {
         if self.active {
-            let rotate_around =
-                self.focus_point.0.unwrap_or_else(Point::origin);
+            let rotate_around: Vector<3> =
+                self.focus_point.0.unwrap_or_else(Point::origin).coords;
```
hannobraun, line 30: Why the explicit type annotation?

```suggestion
            let rotate_around =
                self.focus_point.0.unwrap_or_else(Point::origin).coords;
```

crates/fj-viewer/src/input/rotation.rs
```diff
@@ -26,21 +26,24 @@ impl Rotation {
...
 
             let f = 0.005;
 
             let angle_x = diff_y * f;
             let angle_y = diff_x * f;
 
-            let trans = Translation::from(rotate_around);
+            let trans = Transform::translation(rotate_around);
```
hannobraun, line 37: Again, this is nitpicky, but we're rotation around a point, not a vector. So maybe leave `rotate_around` as a `Point`, and add the `.coords` here, where a vector is needed:
```suggestion
            let trans = Transform::translation(rotate_around.coords);
```

hannobraun (approved):
Thank you, @connor-lennox, this looks great!

I've found one case, where your modifications changed the behavior of the application (mouse wheel input). I've left a comment with a suggestion there. I think I can accept my own suggestion after submitting the review, and merge this immediately.

I've left a bunch of other comment with feedback and suggestions. All of them are optional, and some points (the ones where I noted that I'm being nitpicky), I wouldn't even have mentioned, if you hadn't specifically encouraged the feedback.

One additional note (also nitpicky): Some of the commits could be split into multiple smaller ones, with each smaller commit being focused on one specific change. If some changes were uncontroversial, while others required more discussion, I might want to merge the uncontroversial ones immediately. That would be easier, if every change were its own commit. Doesn't make a difference in this case though.

Feel free to follow up with another pull request that incorporates (some of) my feedback, or not, as you see fit.

crates/fj-viewer/src/graphics/vertices.rs
hannobraun: Hmm, that sounds wrong. `Point` + `Vector` and `Point` - `Vector` should both result in a `Point`, as both operations basically move a point by an offset (the vector). I think you've discovered a bug!

Issue #546, opened 2022-05-06
Inconsistent types in Point arithmetic


connor-lennox:
When adding a Vector to a Point, we receive a Point back:

```rust
impl<V, const D: usize> ops::Add<V> for Point<D>
where
    V: Into<Vector<D>>,
{
    type Output = Self;

    fn add(self, rhs: V) -> Self::Output {
        self.to_na().add(rhs.into().to_na()).into()
    }
}
```

but when *subtracting* a Vector from a Point, we receive a Vector:
```rust
impl<const D: usize> ops::Sub<Self> for Point<D> {
    type Output = Vector<D>;

    fn sub(self, rhs: Self) -> Self::Output {
        self.to_na().sub(rhs.to_na()).into()
    }
}
```

I propose we rewrite the implementation of `Sub` to look more like `Add`, since both operations are effectively translating a `Point` according to a `Vector` (which should, in turn, just give us a different `Point`).

PR #547, opened 2022-05-06
Point arithmetic

connor-lennox:
Addresses #546 .

In addition to adding the new subtraction implementation, I've also gone and applied it to the relevant part of the `vertices.rs` code in a separate commit.


hannobraun (approved):

PR #555, opened 2022-05-09
Fix model README run instructions

chrisprice:
I noticed that the README instructions were generating the following errors for me -

```
% cargo run -- --model cuboid --parameters x=3.0 y=2.0 z=1.0

    Blocking waiting for file lock on build directory
    Finished dev [unoptimized + debuginfo] target(s) in 11.50s
     Running `target/debug/fj-app --model cuboid --parameters x=3.0 y=2.0 z=1.0`
error: Found argument 'y=2.0' which wasn't expected, or isn't valid in this context

USAGE:
    fj-app [OPTIONS]

For more information try --help
```

Looking at the code, I believe this PR updates the instructions to the correct syntax. Certainly the models now load for me -

```
% cargo run -- --model cuboid --parameters x=3.0,y=2.0,z=1.0

    Finished dev [unoptimized + debuginfo] target(s) in 0.30s
     Running `target/debug/fj-app --model cuboid --parameters x=3.0,y=2.0,z=1.0`
  2022-05-09T19:22:00.015053Z  WARN wgpu_core::device: Surface does not support present mode: Mailbox, falling back to FIFO
    at /.../mod.rs:4674

    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
```

### DIFF
--- a/models/cuboid/README.md
+++ b/models/cuboid/README.md
@@ -4,7 +4,7 @@ A model of a simple cuboid that demonstrates sweeping a 3D shape from a primitiv
 
 To display this model, run the following from the repository root (model parameters are optional):
 ``` sh
-cargo run -- --model cuboid --parameters x=3.0 y=2.0 z=1.0
+cargo run -- --model cuboid --parameters x=3.0,y=2.0,z=1.0
 ```
 
 ![Screenshot of the cuboid model](cuboid.png)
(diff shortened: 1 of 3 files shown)


hannobraun (approved):
Thank you @chrisprice! Good catch!

Looks like I made this error once, then copy-pasted it into every new README from then on :smile:

PR #569, opened 2022-05-11
Add local build

### LINKED ISSUE #269: Add local build script
We have a comprehensive CI build now that checks quite a few things. This is good and necessary, but I know from experience, that it can also be a pain for new contributors. You don't know what the CI build checks for yet, so you don't know what to pay attention to. Since CI builds are not instant, it can be a pain to fix your broken build.

Experienced contributors also benefit from an easy way to check the most relevant things locally, before pushing a pull request.

Ideally, this local build would not duplicate what the CI build already does, but due to the way the CI build is structured, this is not practical. And it doesn't matter much, as there are only a few things that are checked. Duplicating those in a local build script is totally fine.

Here's what this build script should do:
- `cargo clippy`
- `cargo test`
- `cargo fmt --check`

This is rather straight-forward. The question is, what should this build script be, as it needs to work on many platforms. @Bandsberg suggested [Nushell](https://www.nushell.sh/) in https://github.com/hannobraun/Fornjot/issues/64#issuecomment-1018751658, but there are many possibilities.

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this is rather self-contained and doesn't require knowledge of Fornjot.

chrisprice:
Fixes #269.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,11 +43,13 @@ CI will kick in automatically when you create a PR. However, to save time you ca
 just build
 ```
 
-You may need to install `just` first:
+You may also need to install `just` first:
 
 ```sh
 cargo install just
 ```
+
+N.B. The justfile is maintained in parallel to the CI configuration. Therefore it's possible they can deviate. Any CI failures not caught by `just build` should be raised as bugs against the justfile. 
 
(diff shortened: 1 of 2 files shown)


hannobraun (approved):
Looks great, @chrisprice, thank you!

PR #570, opened 2022-05-11
Add gnome builder build configuration to .gitignore.

freylint:
Eases fornjot development while using Gnome Builder.

I'm experimenting with this editor and having to be careful not to add this file annoys me.


hannobraun (approved):
Thank you, @freylint!

PR #594, opened 2022-05-16
Support emf

### LINKED ISSUE #154: Export to STL
Fornjot already supports exporting to 3MF. Since STL is also based on triangle meshes, adding support should be fairly straight-forward and not require many changes to existing Fornjot code.

The [export code currently lives in `main.rs`](https://github.com/hannobraun/Fornjot/blob/bf8fe8d05c9b5910366771d76d6890c53d903bad/src/main.rs#L94-L126). It can be extended to check the file ending of the export path, and choose the file format accordingly. I think this would also be a good opportunity to clean up `main.rs` a bit, and move that export code to a separate module.

There might already be libraries for writing STL files that we can use. I haven't looked into that.

Labeling as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this doesn't require much knowledge of Fornjot, and might be a good introduction to it.

chrisprice:
Fixes #154

### DIFF
--- a/crates/fj-export/src/lib.rs
+++ b/crates/fj-export/src/lib.rs
@@ -75,38 +75,49 @@ fn export_stl(mesh: &Mesh<Point<3>>, path: &Path) -> Result<()> {
 
     let vertices = points.iter().map(|points| {
         points.map(|point| {
-            stl_io::Vertex::new([
-                point.x.into_f32(),
-                point.y.into_f32(),
-                point.z.into_f32(),
-            ])
+            [point.x.into_f32(), point.y.into_f32(), point.z.into_f32()]
         })
     });
 
-    let normals = mesh
-        .triangles()
-        .map(|triangle| triangle.points.into())
+    let normals = points
+        .iter()
+        .map(|&points| points.into())
         .map(|triangle: Triangle<3>| triangle.to_parry().normal())
         .collect::<Option<Vec<_>>>()
         .ok_or_else(|| anyhow!("Unable to compute normal"))?;
 
     let normals = normals.iter().map(|vector| vector.into_inner().into()).map(
         |vector: Vector<3>| {
-            stl_io::Normal::new([
+            [
                 vector.x.into_f32(),
                 vector.y.into_f32(),
                 vector.z.into_f32(),
-            ])
+            ]
         },
     );
 
-    let mesh = vertices
+    let triangles = vertices
         .zip(normals)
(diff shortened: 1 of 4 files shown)


crates/fj-export/src/lib.rs
```diff
@@ -14,16 +14,35 @@
 
 #![deny(missing_docs)]
 
-use std::path::Path;
+use std::{fs::File, path::Path};
+
+use anyhow::{anyhow, Result};
```
hannobraun, line 19: Since `fj-export` is a library, [`thiserror`](https://crates.io/crates/thiserror) would be a more appropriate choice here for error handling. Doesn't really matter though. Good enough for now!

crates/fj-export/src/lib.rs
```diff
@@ -48,4 +67,57 @@ pub fn export(mesh: &Mesh<Point<3>>, path: &Path) -> Result<(), Error> {
...
+        .collect::<Option<Vec<_>>>()
+        .ok_or_else(|| anyhow!("Unable to compute normal"))?;
+
+    let normals = normals.iter().map(|vector| vector.into_inner().into()).map(
+        |vector: Vector<3>| {
+            [
+                vector.x.into_f32(),
+                vector.y.into_f32(),
+                vector.z.into_f32(),
+            ]
```
hannobraun, line 95: This kind of code can be shortened a bit:
```suggestion
            vector.components.map(|s| s.into_f32())
```

Not necessary, just noting the possibility.

crates/fj-export/src/lib.rs
```diff
@@ -48,4 +67,57 @@ pub fn export(mesh: &Mesh<Point<3>>, path: &Path) -> Result<(), Error> {
...
+            [point.x.into_f32(), point.y.into_f32(), point.z.into_f32()]
+        })
+    });
+
+    let normals = points
+        .iter()
+        .map(|&points| points.into())
+        .map(|triangle: Triangle<3>| triangle.to_parry().normal())
+        .collect::<Option<Vec<_>>>()
+        .ok_or_else(|| anyhow!("Unable to compute normal"))?;
```
hannobraun, line 87: Since `fj_math::Triangle` is validated on construction, we could add an infallible `normal` method to it, which would make this kind of code simpler. Or, as a more incremental step, thanks to that validation the `anyhow!` here can probably be replaced with an `unreachable!`.

Neither of those need to happen in this pull request though.

hannobraun (approved):
Thank you, @chrisprice, looks excellent!

PR #599, opened 2022-05-17
Improve error handling in `fj_export`

chrisprice:
Thanks for the feedback on #594, I'm always keen to learn. This PR attempts to address the comments. Please feel free to keep the comments flowing!

### DIFF
--- a/crates/fj-export/src/lib.rs
+++ b/crates/fj-export/src/lib.rs
@@ -19,7 +19,7 @@ use std::{fs::File, path::Path};
 use thiserror::Error;
 
 use fj_interop::mesh::Mesh;
-use fj_math::{Point, Triangle, Vector};
+use fj_math::{Point, Triangle};
 
 /// Export the provided mesh to the file at the given path.
 ///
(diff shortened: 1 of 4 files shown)


crates/fj-export/src/lib.rs
```diff
@@ -16,33 +16,33 @@
...
         }
         Some(extension) if extension.to_ascii_uppercase() == "STL" => {
             export_stl(mesh, path)
         }
-        Some(extension) => {
-            Err(anyhow!("Extension not recognised, got {:?}", extension))
-        }
-        None => Err(anyhow!("No extension specified")),
+        Some(extension) => Err(Error::InvalidExtension(
+            extension.to_str().map(|s| s.to_string()),
```
hannobraun, line 39: The `Option<String>` in the error can make for a very confusing error message. How about changing it to `String`, and using something like this here instead:
```suggestion
            extension.to_string_lossy().into_owned(),
```

crates/fj-export/src/lib.rs
```diff
@@ -121,3 +111,27 @@ fn export_stl(mesh: &Mesh<Point<3>>, path: &Path) -> Result<()> {
...
 }
+
+/// An error that can occur while exporting
+#[derive(Debug, Error)]
+pub enum Error {
+    /// No extension specified
+    #[error("no extension specified")]
+    NoExtension,
+
+    /// Unrecognised extension found `{0:?}`
```
hannobraun, line 122: Copy-pasting the error message works for the other doc comments, but not this one :grin: 
```suggestion
    /// Unrecognised extension found
```

hannobraun (approved):
Thank you for the follow-up, @chrisprice!

This looks excellent! I left some comments with suggestions, but this is some high-level complaining I'm doing there. No need to make any changes, but of course, another follow-up PR is welcome, if you're so inclined.

hannobraun:
Hmm, I changed the PR title to highlight the addition of `Triangle::normal`, but on second thought, I'll pull that into a separate PR. `Triangle::normal` and `fj_export::Error` are both additions to the public API. But since they are in different crates, they need to be highlighted in different sections of the Weekly Dev Log and the release changelog.

This is not a complaint about this pull request, I'm just explaining what I'm doing. Feel free to split future pull requests accordingly, if you happen to remember, but otherwise don't worry. Keeping track of this organization stuff is my job :grin:

hannobraun:
Rebased on top of #600. Merging.

PR #604, opened 2022-05-18
Improve Error::InvalidExtension message & docs

chrisprice:
Rounding out the feedback on #599


hannobraun (approved):
Thanks again, @chrisprice! Looks great!

PR #610, opened 2022-05-20
Add serde serialization to `fj`.

freylint:
Adds serde as an optional dependency to `fj`.

I ran into the need this for while marshaling `Shape`s between javascript and rust for the wasm32-unknown-unknown platform.

I believe implementing serde on crates like this is also considered idiomatic. I'm not sure if this implementation is the exact idiomatic way to accomplish that.

### DIFF
--- a/crates/fj/Cargo.toml
+++ b/crates/fj/Cargo.toml
@@ -10,3 +10,10 @@ repository = "https://github.com/hannobraun/fornjot"
 license = "0BSD"
 keywords = ["cad", "programmatic", "code-cad"]
 categories = ["encoding", "mathematics", "rendering"]
+
+[features]
+default = []
+serialization = ["serde"]
+
+[dependencies]
+serde = { version = "1.0.7", optional = true }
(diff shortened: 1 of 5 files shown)


crates/fj/Cargo.toml
```diff
@@ -10,3 +10,10 @@ repository = "https://github.com/hannobraun/fornjot"
 license = "0BSD"
 keywords = ["cad", "programmatic", "code-cad"]
 categories = ["encoding", "mathematics", "rendering"]
+
+[features]
+default = []
+serialization = ["serde"]
```
hannobraun, line 16: I think we don't need an explicit feature here. Just having the implicit `serde` feature should be enough, and is what I've seen in other crates in the Rust ecosystem.

hannobraun (approved):
Thank you, @freylint, look great!

This doesn't seem to have any adverse effect on the compile times, so I'm happy to have it. I've left one comment with a suggestion. Feel free to address that in a follow-up PR or not. I'll merge this either way, once the CI build is green.

Issue #522, opened 2022-05-04
Consider simplifying input/camera approach


hannobraun:
A few months ago, before I started working on Fornjot full-time, I was working on Fornjot's input/camera code, trying to make it more usable than what I've seen in other CAD applications and similar programs. The advanced input/camera system I built includes the following features:

- You can accelerate the zooming speed by continually turning the mouse wheel, and stop an ongoing zoom by turning it into the other direction. The intent here is to allow for both quick movements (to quickly zoom out to get an overview, for example) and fine control while zoomed in closely.
- When moving the model (move mouse while holding right-click), the speed of the movement is dependent on where the mouse pointer is, so the model doesn't move quicker or slower than the mouse moves.
- When rotating the model, it rotates around the point, where the mouse pointer touches the model (called the focus point). The intent here is to allow for close inspection of a feature while zoomed in.

Unfortunately, this never worked very well:
- The zooming is finicky and never worked very well. (no issue exists for this, but #20 suggests improvements)
- Zooming is too slow, if the model is large. (#25)
- The moving behavior works somewhat, but not completely, as the model still moves under the mouse cursor. Maybe the calculation doesn't take perspective into account? (no issue exists for this)
- Moving can be confusing, if you try to move while the mouse doesn't touch the model. (no issue exists for this)
- Rotating the model is completely borked. It can work *somewhat*, but as soon as you move the model, it rotates around totally unpredictable points. (#18, #17 is also relevant)

These problems are still kind of acceptable, as long as Fornjot is still as experimental as it is. But it won't stay acceptable for much longer. There are basically two options:
- Revert to simple and predictable behavior.
- Fix and improve the current behavior.

Or any combination of the two.

I'm labeling this issue as https://github.com/hannobraun/Fornjot/labels/help%20wanted. I don't think I'll have the capacity to focus on this any time soon, so if we are to keep the current non-trivial behavior, someone will have to step up. If no one does, I plan to revert to simple behavior, close all the relevant issues, and add an item about more advanced input to the [feature wishlist](https://github.com/hannobraun/Fornjot/discussions/146).

@ObiWanRohan, you [expressed interest](https://github.com/hannobraun/Fornjot/issues/20#issuecomment-1075518544) in working on input/camera things. Are you still planning to do that?

chrisprice:
I've had a quick dig into this and I'm keen to keep digging. 

[...]

Issue #18, opened 2021-12-04
Computation of focus point is not fully correct


hannobraun:
The focus point is the point on the model (in model coordinates) that the mouse cursor points at. This point is used for input, for example when rotating the model, it is rotated around the focus point.

This works *mostly*, but not fully. The focus point is not detected correctly when the cursor still points at the model, but just barely. Maybe the graphics and the focus point computations use slightly different perspectives, leading to some distortion. This also seems dependent on window size (or at least aspect ratio), as the effect is less pronounced when the window is square.

It might help to implement #17 first, to show what's going on.

hannobraun:
I'd like to add to that that rotation seems a bit weird. It seems to me like it doesn't rotate around the focus point, but a point somewhat offset from that.

Not sure if that was broken from the beginning and I'm just noticing, or if I broke something in the last few days. I'm not really keen to look into it, as long as #17 remains unimplemented.

hannobraun:
@chrisprice fixed the problem I described in the issue description in #614. The problem I described in my follow-up comment (the model not rotating around the focus point) is still present.

PR #614, opened 2022-05-22
Fix x/y field of view conversion

chrisprice:
See https://github.com/hannobraun/Fornjot/issues/522#issuecomment-1133837500


hannobraun (approved):
Thank you for this pull request, @chrisprice!

I can confirm that this pull request fixes a bug, however, it doesn't improve any of the input/camera problems that #522 tracks.

Here's what I found:
- `graphics::Transform::for_vertices` only has one caller, `Renderer::draw`. This is purely graphics output code, so it can't have any influence on the input and camera control code.
- And indeed, the focus point calculation problems (#18) are still there. This can be easily confirmed by moving a model around, then try to rotate it. Once you've moved and rotated it for a bit, it will almost never rotate around the point where the mouse pointer points.

So I wondered, since you didn't include the debug drawing code you mentioned in #522, is there some way I can observe the change made in this PR? Because I didn't really notice a difference under regular circumstances.

Playing around a bit, I noticed a bug that I was previously unaware of: If you make the window narrower (i.e. small in the x direction), the model gets smaller to the point of disappearing, then large to the point of no longer fitting the window, then smaller again. This keeps repeating.

This change fixes that bug: If you make the window narrower, the model just keeps getting smaller, as expected.

So yeah, this is a welcome fix, and I'm definitely merging it. But it seems the bug this fixes has no relation to #18/#522.

chrisprice:
> So I wondered, since you didn't include the debug drawing code you mentioned in https://github.com/hannobraun/Fornjot/issues/522, is there some way I can observe the change made in this PR? Because I didn't really notice a difference under regular circumstances.

Here's a [work in progress version](https://github.com/hannobraun/Fornjot/compare/main...chrisprice:fix-fov) which shows the focus point.

> And indeed, the focus point calculation problems (https://github.com/hannobraun/Fornjot/issues/18) are still there. This can be easily confirmed by moving a model around, then try to rotate it. Once you've moved and rotated it for a bit, it will almost never rotate around the point where the mouse pointer points.

My hunch is that #18 describes at least 2 problems. Quoting from #148 -

> This works mostly, but not fully. The focus point is not detected correctly when the cursor still points at the model, but just barely. Maybe the graphics and the focus point computations use slightly different perspectives, leading to some distortion. This also seems dependent on window size (or at least aspect ratio), as the effect is less pronounced when the window is square.

I believe this is the problem I've fixed.

> I'd like to add to that that rotation seems a bit weird. It seems to me like it doesn't rotate around the focus point, but a point somewhat offset from that.

I believe this is the problem you're still seeing.

Apologies for the lack of clarity. I'm still trying to learn my way around the codebase. I pushed this PR up on its own as it seemed like an easy and obvious standalone fix. 

On reflection I think I've added to the confusion by referencing the PR in #522 instead of explicitly referencing the sub-issue of #18. My goal in referring to #522 was just to show my interest in ultimately tackling that issue. I'll try to be more explicit in future.

hannobraun:
> Here's a [work in progress version](https://github.com/hannobraun/Fornjot/compare/main...chrisprice:fix-fov) which shows the focus point.

Nice! Looking forward to seeing that as a PR :smile: 


> > This works mostly, but not fully. The focus point is not detected correctly when the cursor still points at the model, but just barely. Maybe the graphics and the focus point computations use slightly different perspectives, leading to some distortion. This also seems dependent on window size (or at least aspect ratio), as the effect is less pronounced when the window is square.
> 
> I believe this is the problem I've fixed.

Oh yeah, I totally forgot about that. Even though that's what #18 was about originally, only the "not rotating around focus point" problem stuck in my mind.

And indeed, I can no longer see any deviation here. Moving the model works, if the cursor is exactly over the model, and doesn't work if it's not (as intended).


> > I'd like to add to that that rotation seems a bit weird. It seems to me like it doesn't rotate around the focus point, but a point somewhat offset from that.
> 
> I believe this is the problem you're still seeing.

Yes, I agree.


> Apologies for the lack of clarity. I'm still trying to learn my way around the codebase. I pushed this PR up on its own as it seemed like an easy and obvious standalone fix.

Any incremental improvement is welcome! I just wanted to provide as much context as I could, for your benefit, and the benefit of anyone who's going to read this later (including myself). This wasn't meant as criticism for not fixing the problem completely, or anything like that.


> On reflection I think I've added to the confusion by referencing the PR in #522 instead of explicitly referencing the sub-issue of #18. My goal in referring to #522 was just to show my interest in ultimately tackling that issue. I'll try to be more explicit in future.

Well, as you noted, #18 referred to 2 different problems in the first place, so if there was any confusion, that was the source.

Issue #263, opened 2022-02-28
Use static typing to distinguish between degrees and radians


hannobraun:
Angles in the `fj` library are just represented as floating-point numbers in radians. This is error-prone, as users can assume those are degrees. 

I think there should a an `Angle` type that uses radians internally, but provides constructors (`from_deg`, `from_rad`) and accessors (`deg`, `rad`) to make working with angles convenient and obvious. This type should also implement the common math operators.

Such a type will probably be useful in the CAD kernel too, so #122 is relevant.

hannobraun:
This is a rather self-contained change that could be a good introduction to the `fj` crate, so I'm labeling it as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue.

Here's what would need to happen to address this, basically:

1. Create an `Angle` struct, basically a wrapper around `f64`, and add it to the `fj` crate. Not sure where the best place would be, but just adding it to the root module, or a new `angle` module would be perfectly acceptable. We can always adjust later.
2. Add the aforementioned `from_deg`/`from_rad`/`deg`/`rad` methods. Implementing operators like `std::ops::Add`/`Sub`/... would be nice, but can be left for later.
3. Review the whole API of `fj`, and use the new `Angle` type everywhere to represent angles.

Ideally, we'd add an `Angle` type to `fj-math`, and just re-export that from `fj`. But as long as we haven't figured out #122, that's not an option. And in the meantime, doing what I described here is certainly better than the current situation.

PR #619, opened 2022-05-23
Create and use `Angle` type

gabsi26:
This PR creates and uses a new `Angle` type in the `fj` module as suggested by #263

### DIFF
--- a/crates/fj/src/angle.rs
+++ b/crates/fj/src/angle.rs
@@ -17,7 +17,7 @@ impl Angle {
     }
     /// Create a new angle specified in degrees
     pub fn from_deg(deg: f64) -> Self {
-        Self::from_rad(deg * PI / 180.)
+        Self::from_rad(deg.to_radians())
     }
     /// Retrieve value of angle as radians
     pub fn rad(&self) -> f64 {
@@ -25,7 +25,7 @@ impl Angle {
     }
     /// Retrieve value of angle as degrees
     pub fn deg(&self) -> f64 {
-        self.rad / PI * 180.
+        self.rad.to_degrees()
     }
 
     // ensures that the angle is always 0 <= a < 2*pi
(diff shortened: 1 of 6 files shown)


crates/fj/src/angle.rs
```diff
@@ -0,0 +1,101 @@
...
+impl Angle {
+    /// Create a new angle specified in radians
+    pub fn from_rad(rad: f64) -> Self {
+        Self {
+            rad: Self::wrap(rad),
+        }
+    }
+    /// Create a new angle specified in degrees
+    pub fn from_deg(deg: f64) -> Self {
+        Self::from_rad(deg * PI / 180.)
```
hannobraun, line 20: Can be simplified using [`f64::to_radians`](https://doc.rust-lang.org/std/primitive.f64.html#method.to_radians):
```suggestion
        Self::from_rad(deg.to_radians())
```

crates/fj/src/angle.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+    /// Retrieve value of angle as radians
+    pub fn rad(&self) -> f64 {
+        self.rad
+    }
+    /// Retrieve value of angle as degrees
+    pub fn deg(&self) -> f64 {
+        self.rad / PI * 180.
```
hannobraun, line 28: Can be simplified using [`f64::to_degrees`](https://doc.rust-lang.org/std/primitive.f64.html#method.to_degrees):
```suggestion
        self.rad.to_degrees()
```

hannobraun (dismissed):
Thank you, @gabsi26, this looks awesome!

I've left two minor suggestions. I can probably accept those myself after submitting this review.

hannobraun:
Added a commit to apply my suggestions. Also rebased on `main` (I've been merging two of my own pull requests while reviewing this one). Will merge, once the CI build is green.

Thanks again, @gabsi26!

PR #621, opened 2022-05-23
Fix mistake in `Angle::wrap`

gabsi26:
Sorry for the inconvenience, but I made a mistake in the wrapping function, where it previously simply flipped the sign of a negative angle. Now it should properly "overflow" from 2*pi

### DIFF
--- a/crates/fj/src/angle.rs
+++ b/crates/fj/src/angle.rs
@@ -32,7 +32,7 @@ impl Angle {
     fn wrap(rad: f64) -> f64 {
         let modulo = rad % (2. * PI);
         if modulo < 0. {
-            modulo * -1.
+            2. * PI + modulo
         } else {
             modulo
         }


hannobraun (approved):
Thank you, @gabsi26!

> Sorry for the inconvenience

Don't worry about it! I didn't notice it either. I looked right at that code during review, and my brain just went like "looks legit" :smile: 

Mistakes happen, and I'm glad you noticed and fixed this one. No need to apologize for that!

hannobraun:
@gabsi26 You can run `cargo fmt` locally, which should just take care of the formatting automatically. This requires `rustfmt` to be installed, which should have happened automatically, due to our `rust-toolchain.toml` file.

hannobraun:
Ran `cargo fmt` and squashed all the commits. Should be good to go now.

Thanks again, @gabsi26!

PR #623, opened 2022-05-24
Fix bounding volume issue for swept shapes

### LINKED ISSUE #566: Bounding volume of sweep operation is incorrect
Sweeps have recently been made more flexible. Now it's possible to sweep along arbitrary straight paths, not just along the z-axis. This seems to have introduced a bug in the AABB calculation.

Screenshot from 2022-05-11 15-57-16

As you can see here, the near plane slices through the model, showing us the inside. This shouldn't happen at this camera distance.

I'm pretty sure that this piece of code needs to be updated:
https://github.com/hannobraun/Fornjot/blob/3d5500e95621ebe5c786a886a11eaaaab04bd618/crates/fj-operations/src/sweep.rs#L24-L27

Including the target point to extend the original bounding box is not sufficient. I think we instead need to created a transformed copy (moved by `self.path`) of the bounding box, then merge both bounding boxes.

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this most likely just requires a focused change to the code I quoted above.

gabsi26:
This addresses #566

This solution is probably less than ideal,
since the bounding volume gets split up into its vertices,
which then get individually translated.

Signed-off-by: gabsi26 <gabriel.seebach@gmx.at>

edit (by @hannobraun):
Close #566 (to make sure the issue is closed automatically when this is merged)

### DIFF
--- a/crates/fj-operations/src/sweep.rs
+++ b/crates/fj-operations/src/sweep.rs
@@ -24,6 +24,13 @@ impl ToShape for fj::Sweep {
 
     fn bounding_volume(&self) -> Aabb<3> {
         let target = Point::origin() + self.path();
-        self.shape().bounding_volume().include_point(&target)
+        self.shape()
+            .bounding_volume()
+            .merged(&Aabb::<3>::from_points(
+                self.shape()
+                    .bounding_volume()
+                    .vertices()
+                    .map(|v| v + self.path()),
+            ))
     }
 }
--- a/crates/fj-operations/src/sweep.rs
+++ b/crates/fj-operations/src/sweep.rs
@@ -23,7 +23,6 @@ impl ToShape for fj::Sweep {
     }
 
     fn bounding_volume(&self) -> Aabb<3> {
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you, @gabsi26! I'll do a proper review later, but this looks good to me so far. I'll write up some more thoughts in #566.

crates/fj-operations/src/sweep.rs
```diff
@@ -24,6 +24,13 @@ impl ToShape for fj::Sweep {
 
     fn bounding_volume(&self) -> Aabb<3> {
         let target = Point::origin() + self.path();
```
hannobraun, line 26: This variable is no longer used. Let's remove it.
```suggestion
```

hannobraun (commented):
Thank you, @gabsi26, looks and works excellently!

I left one suggestion, which I can probably apply myself, after I submit this pull request.

PR #627, opened 2022-05-24
Replace problematic `Edge` constructor with a simpler one

hannobraun:
The constructor was problematic, because it projected the edge vertices into the curve to create the local form of the vertices, without checking whether those points lie on the curve. I also had some other concerns regarding floating point accuracy issues in the computation of these surface coordinates, but I'm not sure whether those concerns where well-founded.

In any case, the new constructors accepts the canonical and the local forms of the vertices. It doesn't verify them directly, but geometric validation will later ensure that they match.


hannobraun:
A change in this pull request triggers a geometric validation issue in the star model, hence the CI failure. I've been working on improving the error reporting infrastructure, to help track this down.

PR #628, opened 2022-05-24
Correct face orientation fixes #506

### LINKED ISSUE #506: Invalid face orientation when sweeping in the negative direction
As of #505, sweeping in arbitrary directions is supported. This can lead to face orientation issues, when sweeping downwards:
Screenshot from 2022-04-26 15-11-29

The best way to address this is probably to catch that case in the kernel, and invert the surfaces accordingly.

gabsi26:
This PR simply checks if the path of a sweep is pointing in a negative direction.
The change is quite minimal and might not be sufficient for all cases

### DIFF
--- a/crates/fj-kernel/src/algorithms/sweep.rs
+++ b/crates/fj-kernel/src/algorithms/sweep.rs
@@ -20,12 +20,18 @@ pub fn sweep_shape(
     let translation = Transform::translation(path);
 
     let (mut bottom, source_to_bottom) = source.clone_shape();
-    bottom
-        .update()
-        .update_all(|surface: &mut Surface| *surface = surface.reverse())
-        .validate()?;
-
     let (mut top, source_to_top) = source.clone_shape();
+
+    if path.dot(&Vector::from([0., 0., 1.])) >= fj_math::Scalar::from_f64(0.) {
+        bottom
+            .update()
+            .update_all(|surface: &mut Surface| *surface = surface.reverse())
+            .validate()?;
+    } else {
+        top.update()
+            .update_all(|surface: &mut Surface| *surface = surface.reverse())
+            .validate()?;
+    }
     transform_shape(&mut top, &translation);
 
     let mut target = Shape::new();
--- a/crates/fj-kernel/src/algorithms/sweep.rs
+++ b/crates/fj-kernel/src/algorithms/sweep.rs
@@ -124,11 +124,22 @@ pub fn sweep_shape(
                 let top_edge =
                     source_to_top.edges().get(&edge_source).unwrap().clone();
 
-                let surface =
+                let surface = if path.dot(&Vector::from([0., 0., 1.]))
+                    >= fj_math::Scalar::from_f64(0.)
+                {
                     target.insert(Surface::SweptCurve(SweptCurve {
                         curve: bottom_edge.get().curve(),
                         path,
-                    }))?;
+                    }))?
+                } else {
+                    target.insert(
+                        Surface::SweptCurve(SweptCurve {
+                            curve: bottom_edge.get().curve(),
+                            path,
+                        })
+                        .reverse(), ////////////////////////////////////
+                    )?
+                };
 
                 let cycle = target.merge(Cycle::new(vec![
                     bottom_edge,


hannobraun (changes requested):
Thank you for the pull request, @gabsi26!

First off, I found that switching to an opaque color made it way easier to figure out what's going on. Maybe we should switch the model colors back to something simple, and just have one transparent model for testing, but that has nothing to do with this pull request.

This pull request addressed only part of #506, by fixing the top and bottom faces. It does not address the side faces. See here:
Screenshot from 2022-05-24 16-41-38

(That's just the star model, except `h` in line 59 was replaced with `-h`.)

I'd be happy to merge a partial fix, but the commit message of your first commit contains the magic incantation that will make GitHub close the issue upon merging this. Could you please either remove the reference to the issue from the commit message (straight-forward to do using something like `git rebase -i main`, in case you didn't know), or fix the side walls?

gabsi26:
Well i tried getting rid of the reference but it doesn't seem to work.
I'm not that fluent in guit unfortunatelly

hannobraun:
> Well i tried getting rid of the reference but it doesn't seem to work.
> I'm not that fluent in guit unfortunatelly

I know that Git can be hard to deal with. Luckily (for me), it gets easier after 10+ years :smile: 

I'll take care of it. I'll post the steps I'm taking here, in case you're interested.

gabsi26:
Yes please that would be great as a learning resource :D

In the meantime I think i really fixed the issue^^

hannobraun:
Oh man, that branch was truly broken :smile: 

Not sure what happened, but as a general note, `git merge` is often not a great option in pull requests, as it creates a complicated history. I suspect that in this case, there was some back-and-forth merging going on, seeing how confused the branch was.

In general, if your pull request is no longer up-to-date, the best way to address that is `git rebase`, which just takes your commits and replays them on top of another branch. If any merge conflict arises, you can fix them commit by commit. This leaves behind a linear history, without any merge commits with multiple parents.

Let's say, you're working on the branch `my-fix`, and have just pulled the latest changes from the upstream repository into `main`. Go to your branch (`git switch my-fix`) and run `git rebase main`, fix each conflict (if there are any). Once you fixed a conflict, use `git rebase --continue` to go to the next one. You can also use `git rebase --skip` to throw away one of your commits, or `git rebase --abort`, if you're in over your head and just want to go to where you were before you started rebasing.

---

Okay, but enough introduction. Here's what I did to fix the branch.

First, I tried a `git rebase main` to clean up the history. One complication was that Git tried to apply the same commit twice. Not sure why, but given how confusing the branch was, I'm not surprised Git does confusing things with it. When Git showed me the commit, I verified (using `git log`) that the commit it complained about was already applied, and used `git rebase --skip` to throw it away.

Next, I looked at the branch (`git log`) and saw it still contained lots of commits that didn't belong there. Again, not sure why. Fortunately, in this case it was pretty easy to distinguish between commits from your branch, and commits I've made myself earlier today. I decided to sort through them manually.

So, `git rebase -i main` (`-i` for *interactive*), and tell Git which ones I'd like to get rid of:
Screenshot from 2022-05-24 17-43-46

You can just remove the lines of the commits you want to drop instead of writing `drop`, but I figured this way it's easier to see what I did.

Saved the file, quit the editor, and Git went ahead.

Screenshot from 2022-05-24 17-44-50

`git log` now looks good.

Screenshot from 2022-05-24 17-45-26

I noticed at this point, that the reference to the issue was gone. Looks like your attempt to update the commit was successful in that regard. If it hadn't been gone, I'd have run another `git rebase -i main`, and marked the commit with `r` (for `reword`) instead of `pick`. Git should then take care of the rest, showing you an editor where you can edit the commit message.

Now I wanted to include your new commit, but didn't want to start over. So first, I told Git to get the latest version of your branch, without merging it into my local branch using `git fetch`.

Screenshot from 2022-05-24 17-49-17

You can see the reference to `FETCH_HEAD` here, which is a convenient shortcut for "whatever the thing is you just fetched". So now I can do a `git log FETCH_HEAD`.
Screenshot from 2022-05-24 17-50-50

Copy the commit id of your new commit, and get it into my local branch using `git cherry-pick`.
Screenshot from 2022-05-24 17-51-40

`git cherry-pick` is a bit like `git merge`, except it only merges exactly the commits you tell it to, not their history. So no risk of messing anything up again.

Now the branch looks good to me.
Screenshot from 2022-05-24 17-52-50

Next, force-push the repaired branch to GitHub (`git push -f`), and here it is. Refresh this page, and GitHub only shows those 3 commits.

Last step: Hope I didn't miss anything, because digging through forgotten history using `git reflog` is where the pain really starts :smile:

crates/fj-kernel/src/algorithms/sweep.rs
```diff
@@ -124,11 +124,22 @@ pub fn sweep_shape(
                 let top_edge =
                     source_to_top.edges().get(&edge_source).unwrap().clone();
 
-                let surface =
+                let surface = if path.dot(&Vector::from([0., 0., 1.]))
```
hannobraun, line 127: Would be nicer, if that check happened only once, and the result was stored into a variable with a descriptive name (`sweep_along_negative_direction`, or something like that).

With two separate checks, it becomes easy to miss one, once they need to change.

crates/fj-kernel/src/algorithms/sweep.rs
```diff
@@ -124,11 +124,22 @@ pub fn sweep_shape(
...
+                    >= fj_math::Scalar::from_f64(0.)
```
hannobraun, line 128: Note that `Scalar` has some short-hands for often-used values, in this case `Scalar::ZERO`.

gabsi26:
Thanks a lot for this indepth explanation and cleaning up my mess. I hope I won't cause something like this again, but at least nowthere is a guide how to fix it :D

crates/fj-kernel/src/algorithms/sweep.rs
```diff
@@ -124,11 +124,22 @@ pub fn sweep_shape(
...
                         path,
-                    }))?;
+                    }))?
+                } else {
+                    target.insert(
+                        Surface::SweptCurve(SweptCurve {
+                            curve: bottom_edge.get().curve(),
+                            path,
+                        })
+                        .reverse(), ////////////////////////////////////
```
hannobraun, line 140: Would be a bit nicer to reduce some duplication here. Something like this:

``` rust
let mut surface = Surface::SweptCurve(...);

if sweep_along_negative_direction {
    surface = surface.reverse();
}

let surface = target.insert(surface)?;
```

hannobraun:
> Thanks a lot for this indepth explanation and cleaning up my mess. I hope I won't cause something like this again, but at least nowthere is a guide how to fix it :D

You're welcome! And don't worry, it happens. Git can be gnarly, but most of these problems are relatively easy to fix, once you understand what's happening and know enough Git commands to clean it up. Just takes time.

hannobraun (approved):
Thank again for this pull request, @gabsi26!

It looks good to me now. I've left some suggestions for improvements, but it's fine. Those things can be addressed at some later point. Feel free to submit a follow-up PR, if you're up for it, or just leave it, as you prefer.

Merging now.

PR #633, opened 2022-05-25
Improve error handling in `fj-viewer`

hannobraun:
Shape processing errors will no longer cause a panic.


hannobraun:
Found some more improvements to make, that I want to include in this pull request. Will push more commits shortly.

PR #641, opened 2022-05-26
Improve `Angle`

T0mstone:
This is a suggestion for a slight improvement to the `Angle` type.

1. It replaces the usage of `2. * PI` with `TAU`: This constant exists, so why not use it?
2. It adds two new units that an angle can be converted to/from. I think gon is used in engineering, so it could prove useful; Revolutions is just good to have in any case (say you want to rotate something 1/3 of a turn, now it's super easy)
3. It removes the multiplication of two angles. Multiplying two angles is mathematically only useful if talking about [solid angles](https://en.wikipedia.org/wiki/Solid_angle), which this definetely wasn't. I don't see a context in which you would multiply an angle by another angle. Rather, you'll probably want to multiply an angle by a simple factor (something like "2 times that angle" or "1/3 of that angle"), so I implemented that. Dividing an angle by an angle is well-defined, so I kept it.

I don't think the changed API was used anywhere yet, but I'm not familiar with the codebase. This just jumped out at me when reading the PR that introduces angles, so I thought to just try a PR of my own.


hannobraun (approved):
Thank you for the pull request, @T0mstone, these are some great changes!

Issue #72, opened 2022-01-21
Make defining models more convenient


hannobraun:
The `extern "C"` stuff required for the dynamic library isn't great. And the way parameters are passed to the model has caused further pain. Ideally, creating a model should require little or no ceremony to satisfy the plugin system. Maybe something like this would be good:

``` rust
/// Defines the model parameters
///
/// The `derive` makes the parameter parsing just work, so the user doesn't have
/// to think about it. Preferably in a way that moves the bulk of the work to
/// the host, so compile times for the model stay low.
#[derive(fj::Parameters)]
pub struct Spacer {
    pub outer: f64,
    pub inner: f64,
    pub height: f64,
}

impl fj::Model for Spacer {
    /// Defines the model itself
    /// 
    /// Has access to the parameters through `self`. Method is defined in a
    /// trait, to provide type safety.
    fn model(&self) -> fj::Shape {
        // define the model
    }
}
```

This is just a sketch, not to be taken too seriously. Maybe it's possible, or maybe a `#[no_mangle]` or something will still be required.

mdaffin:
If you use proc_macros you can make them a single function with arguments and expand that into whatever you need to meet the C interface required.

```rust
#[fj::model]
pub fn model(outer: f64, inner: f64, height: f64) -> fj::Shape { /* ... */ }
```

A bit more effort to define the proc_macro, but allows you to change the underlying implementation in the future if required. And keeps a nice simple interface for users to use with minimal boiler plate.

hannobraun:
Thanks for the suggestion, @mdaffin!

Yes, something like that should be possible. The only thing I'm not sure about, is if proc macros can add attributes, so maybe something like `#[no_mangle]` would still be in there. We'll see one someone gets a chance to explore this more closely.

I actually like your suggested syntax better than my own! It's just simpler and cleaner.

mdaffin:
Proc macros can do anything. They just take in a token stream and output a possibly different token stream. For instance, the test_case proc macro adds the `#[test]` attribute the the tests it generates: https://github.com/frondeus/test-case/blob/master/src/test_case.rs#L78

hannobraun:
Ah, good to know, thanks!

I've used proc macros before, but never to add attributes. I had this worry in the back of my mind about order of execution in the compiler, because proc macros often are attribute, so can they add something that's already been read at point in compilation. But good to know my worries were unfounded!

hannobraun:
[...]

PR #643, opened 2022-05-29
Provides two macros to make model creation more convenient

gabsi26:
An initial PR to get this rolling.

This PR deals with #72. Where `attributed_arguments` is a crude implementation of an idea https://github.com/hannobraun/Fornjot/issues/72#issuecomment-1138387203

The attribute `#[value()]` is required for each parameter the user wants to pass to the model it accepts different forms
- `#[value(3)]` sets default value to 3
- `#[value(default=3)]` same as above
min and max can additionally be supplied like this:
`#[value(min=1, max=2)]`
As long as either one of the three values is supplied a default will be set, min and max are currently only used if no default is provided to set it. Where min takes precedence over max.
Later on a doc comment containing the provided min and max values should be introduced.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ members = [
     "crates/fj-math",
     "crates/fj-operations",
     "crates/fj-viewer",
+    "crates/fj-proc",
 
     "crates/fj-window",
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ members = [
     "models/cuboid",
     "models/spacer",
     "models/star",
+    "models/star_attributed_arguments",
     "models/test",
 
     "tools/export-validator",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,11 @@ members = [
     "crates/fj-math",
     "crates/fj-operations",
     "crates/fj-viewer",
+
     "crates/fj-window",
+
+    "crates/fj-proc",
+
 
     "models/cuboid",
     "models/spacer",
--- a/crates/fj-proc/src/attributed_arguments.rs
+++ b/crates/fj-proc/src/attributed_arguments.rs
@@ -4,14 +4,13 @@ use syn::{
     bracketed, parenthesized, parse::Parse, parse_macro_input, parse_quote,
 };
 
-pub fn attributed_arguments(
-    _default_values: TokenStream,
-    input: TokenStream,
-) -> TokenStream {
+pub fn attributed_arguments(_: TokenStream, input: TokenStream) -> TokenStream {
     let item = parse_macro_input!(input as syn::ItemFn);
     let inputs = item.clone().sig.inputs;
+
     let args: Vec<Argument> =
         inputs.iter().map(|inp| parse_quote!(#inp)).collect();
+
     let mut defaults = Vec::new();
     let mut mins = Vec::new();
     let mut maxs = Vec::new();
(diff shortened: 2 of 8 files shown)


hannobraun:
Thank you for the pull request, @gabsi26! I'll take a look later, possibly tomorrow.

Cargo.toml
```diff
@@ -10,11 +10,17 @@ members = [
     "crates/fj-math",
     "crates/fj-operations",
     "crates/fj-viewer",
+    "crates/fj-proc",
+
```
hannobraun, line 14: The rest of that list is alphabetically sorted. Would be nice to keep that. Also, no need for an empty line.

Cargo.toml
```diff
@@ -10,11 +10,17 @@ members = [
...
     "crates/fj-window",
 
+    "crates/fj-proc",
+
+
```
hannobraun, line 19: There's the same crate a second time. I'm surprised that doesn't result in some kind of error! :smile:

models/star/Cargo.toml
```diff
@@ -8,3 +8,6 @@ crate-type = ["cdylib"]
 
 [dependencies.fj]
 path = "../../crates/fj"
+
+[dependencies.fj-proc]
+path = "../../crates/fj-proc"
```
hannobraun, line 13: I think it should be possible, to re-export `fj-proc` from `fj` (`pub use fj_proc;`). Then we wouldn't need the extra dependency, and would also have a simpler syntax in the models (`fj::model` instead of `fj_proc::model`).

crates/fj-proc/src/attributed_arguments.rs
```diff
@@ -0,0 +1,198 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{
+    bracketed, parenthesized, parse::Parse, parse_macro_input, parse_quote,
+};
+
+pub fn attributed_arguments(_: TokenStream, input: TokenStream) -> TokenStream {
```
hannobraun, line 7: I think having two different attributes (and two different models for testing/demonstrating) them is overkill. I like the approach implemented in this one much better. It would be better, if this just replaced the other one.

crates/fj-proc/src/attributed_arguments.rs
```diff
@@ -0,0 +1,198 @@
...
+        for value in arg.attr.values.clone() {
+            if let Some(ident) = value.ident.clone() {
+                match ident.to_string().as_str() {
+                    "default" => default = Some(value.val.clone()),
+                    "min" => min = Some(value.val.clone()),
+                    "max" => max = Some(value.val.clone()),
+                    _ => {}
+                }
+            } else {
+                default = Some(value.val.clone());
```
hannobraun, line 34: Maybe it would be better not to have this fallback, and always require an explicit `default = ` to specify default values. Would be more verbose in the simple case, but also easier to read and understand.

What do you think?

crates/fj-proc/src/attributed_arguments.rs
```diff
@@ -0,0 +1,198 @@
...
+                #block
+            }
+        }.into()
+}
+
+/// Represents one parameter given to the `model`
+/// `#[value(default=3, min=4)] num_points: u64`
+/// `^^^^^^^^^^^^^^^^^^^^^^^^^^ ~~~~~~~~~~  ^^^-- ty`
+/// `           |                    |`
+/// `         attr                 ident`
```
hannobraun, line 59: Nice style of documentation :+1:

hannobraun (approved):
Thank you again for the pull request, @gabsi26! This is an awesome improvement!

I left a bunch of comments with specific suggestions, as well as more general remarks on where I'd like to see things going. Feel free to address those in a follow-up PR, if you're up for it. I'll merge this one now, as it provides an excellent base for further iteration.

crates/fj-proc/src/attributed_arguments.rs
gabsi26: The first one was more or less just left in because it was the first one I came up with. So more for completeness than anything else. I'll get rid of it :)

hannobraun:
One additional note: I just discovered, that `Cargo.lock` wasn't up-to-date, meaning Cargo will immediately update it on any `cargo build`, `cargo test`, or similar. Then there will be a local change, which can confuse contributors and could interfere with work in the Git history (like somebody running a `git bisect`).

No big deal, I'll fix it (I'm about to do my weekly `cargo update` anyway). But in general, it would be nicer if changes to `Cargo.lock` were committed in the same commit that triggered them (like in this case, the commit that adds the new crate).

crates/fj-proc/src/attributed_arguments.rs
gabsi26: I think I'll simplify the macro to not allow for this implicit fallback

PR #644, opened 2022-05-29
Fix camera rotation around focus point

chrisprice:
This fixes the remaining part of #18.

I've noticed a couple of possible further issues -

* The larger the rotation, the less intuitive the axis of rotation becomes. This is due to it being fixed to x/y axis rather than relative to `Camera` `rotation`.
* Whilst the transform in `Camera` is broken into `rotation` and `translation` matrices, the current implementation of `Rotation` does not guarantee `rotation` is an [orthogonal matrix (under heading Properties of the Transform Matrix)](http://fastgraph.com/makegames/3drotation/).

### DIFF
--- a/crates/fj-viewer/src/input/rotation.rs
+++ b/crates/fj-viewer/src/input/rotation.rs
@@ -43,7 +43,7 @@ impl Rotation {
 
             let inv = trans.inverse();
 
-            camera.rotation = trans * rot_y * rot_x * inv * camera.rotation;
+            camera.rotation = camera.rotation * trans * rot_y * rot_x * inv;
         }
     }
 }


hannobraun (approved):
Thank you @chrisprice, looks and works great!

> The larger the rotation, the less intuitive the axis of rotation becomes. This is due to it being fixed to x/y axis rather than relative to `Camera` `rotation`.

Good point! Feel free to open an issue about that. If not, I can do it myself (already made a note, will do it once I get to it).

> Whilst the transform in `Camera` is broken into `rotation` and `translation` matrices, the current implementation of `Rotation` does not guarantee `rotation` is an [orthogonal matrix (under heading Properties of the Transform Matrix)](http://fastgraph.com/makegames/3drotation/).

That looks like an interesting article! Unfortunately I don't have the time to read it completely right now, so I don't quite follow why this lack of orthogonality is a problem. But I'd just like to emphasize that the way that code is written isn't very intentional (beyond trying to get a certain result), so further improvements are certainly welcome.

PR #646, opened 2022-05-30
fj: fix memory leak in Sketch

### LINKED ISSUE #186: `fj::Sketch` is leaking memory
If an `fj::Sketch` is dropped, it will leak the memory of the `Vec` that was provided when constructing it.

I see two solutions for addressing this:
1. Add a `Drop` implementation that frees the memory. Careful though: Currently, `Clone` is derived, which we can get away with, because the raw `Vec` isn't modified, and ownership isn't given away. With a ` Drop` implementation, we also need a custom `Clone` implementation that clones the raw `Vec`. Otherwise, `fj::Sketch` will become unsound.
2. I've recently become aware of [`abi_stable`](https://crates.io/crates/abi_stable). I don't fully understand what it does, but I do know that it contains an [FFI-safe alternative to `Vec`](https://docs.rs/abi_stable/0.10.4/abi_stable/std_types/struct.RVec.html). By using that, we could get rid of the `unsafe` and thus the potential for unsoundness.

I haven't evaluated `abi_stable` yet, but unless there's a reason not to use it, I think it's the preferable option.

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this is fairly self-contained and can serve as an introduction to the `fj` library.

Thanks to @WeirdConstructor, who reported this [on Matrix](https://matrix.to/#/#fornjot:braun-odw.eu).

kamirr:
Fixes #186

### DIFF
--- a/crates/fj/src/shape_2d.rs
+++ b/crates/fj/src/shape_2d.rs
@@ -1,4 +1,5 @@
 use std::mem;
+use std::sync::atomic;
 
 use crate::Shape;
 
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @kamirr, looks great!

PR #652, opened 2022-05-31
Proc me up again

gabsi26:
This is a follow-up PR to #643 dealing with some of the suggestions provided there.

Now the possible annotations for an argument are permutations of `default`, `min`, `max` with `=` and a value as well as empty or no parentheses.

If supplied min and max values are automatically checked.
No implicit determination of default values takes place anymore


hannobraun (approved):
Thank you, @gabsi26, looks great!

gabsi26:
Currently if a user decides not ot supply any metadata an annotation like `#[value]` is still required in front of the argument.
Do you think this requirement should be removed? @hannobraun

PR #669, opened 2022-06-05
Rotate around axis relative to camera

### LINKED ISSUE #658: Rotation axis should be more intuitive
As noted by @chrisprice in #644, the camera rotation can be unintuitive, as the model rotates around a fixed axis, not an axis relative to the camera. This can be reproduced somewhat easily using the test model, by starting a rotation and moving the mouse in a few different directions. It is quite easy to get into a situation, where the direction of the mouse movement no longer has any relation to the direction of the rotation.

It would probably be best to always rotate around an axis relative to the camera, so the rotation always follows the movement of the mouse.

chrisprice:
Fixes #658.

Disclaimers 😄 

* I've opted for a relatively lightweight change. I can think of more intrusive changes which extend the types around `Transform` but didn't want to jump into that without discussion. 
* I'm also not an expert in this area, I've just been following through [this article](http://fastgraph.com/makegames/3drotation/). There are likely to be mathematical simplifications that I've missed.
* I've noticed discussion about removing the dependency on `nalgebra` so I haven't dug into its API looking for appropriate implementations of the logic I've added. There's very likely to be something in there that I've missed.

### DIFF
--- a/crates/fj-math/src/transform.rs
+++ b/crates/fj-math/src/transform.rs
@@ -131,6 +131,18 @@ impl Transform {
     pub fn data(&self) -> &[f64] {
         self.0.matrix().data.as_slice()
     }
+
+    /// Extract the rotation component of this transform
+    pub fn extract_rotation(&self) -> Transform {
+        Self(nalgebra::Transform::from_matrix_unchecked(
+            self.0.matrix().fixed_resize::<3, 3>(0.).to_homogeneous(),
+        ))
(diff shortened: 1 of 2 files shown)


hannobraun (approved):
Thank you, @chrisprice, this works great!

Sorry for the delay in reviewing this. Monday was a public holiday here in Germany, and I spent most of it away from the computer.

Issue #684, opened 2022-06-12
Some crates are missing from `default-members` in `Cargo.lock`


hannobraun:
In the top-level `Cargo.lock` file, not all crates from the `crates/` directory are present in `default-members/`. This means that parts of the CI build (and the `justfile`) don't apply to them. Running `cargo test` from the repository root, for example, will not run tests from these crates.

This should be addressed. Everything from `crates/` should be in `default-members`. Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, since this is a small change that doesn't require any specific knowledge about Fornjot.

Issue #683, opened 2022-06-12
CI build should enable all features


hannobraun:
Right now, the CI build doesn't do that, which let errors slip in. See #681.

Hopefully passing `--all-features` to all the `cargo build`/`cargo test`/`cargo clippy` invocations will be enough.

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, since addressing this requires no knowledge on Fornjot specifics.

PR #682, opened 2022-06-11
Fix fj serialization

kamirr:
Fixes #681.

Two things I'm unsure about:
1. `just build` doesn't run the tests I've added to `fj`, despite them running fine via `cargo test`. Am I missing something?
2. Should I commit changes to `Cargo.lock`?

PR #685, opened 2022-06-13
Followup on PR #682

kamirr:
1. Address the ordering of dependencies in `fj`
2. Add `fj` and `fj-proc` to `default-members` so that `just build` takes them into account
3. Make sure that the tests use `--all-features`

Besides cleaning up #682, this addresses #683 and #684.

I've verified the changes, other than the CI. I'll check the results for this PR to make sure that it works as intended.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,7 +116,9 @@ jobs:
         uses: actions-rs/cargo@4ff6ec2846f6e7217c1a9b0b503506665f134c4b
         with:
           command: build
+          args: --all-features
       - name: Run `cargo test`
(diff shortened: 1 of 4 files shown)


kamirr:
Ok, looking at the CI results it appears that `cargo build`/`test`/`clippy` are all invoked with the `--all-features` flag as intended.

kamirr:
Ok, looking at the CI results it appears that `cargo build`/`test`/`clippy` are all invoked with the `--all-features` flag as intended.

hannobraun (approved):
Thank you @kamirr, looks great!

PR #692, opened 2022-06-13
Improve --parameters usability

A-Walrus:
The example command in the readme didn't work so I fixed it. Also added trimming of white-space to allow spaces before/after commas and equal signs.


hannobraun (approved):
Thank you, @A-Walrus, nice changes!

Issue #697, opened 2022-06-18
Try to abolish `Shape`


hannobraun:
`Shape` is the central data structure that `fj-kernel` uses to store the objects that make up a shape. It was initially introduced for the following reasons:

- **Validation:** Storing objects in a central location gives you a point where you can implement validation logic. I'm confident that there are solutions to achieve validation without `Shape` though (see #696).
- **Object identity:** An initial idea was to create objects once, and only refer to the existing object from then on, instead of allowing identical objects to be created. This idea turned out to be not that necessary, given that the validation infrastructure already prevents a lot of errors, and was later largely sacrificed for the sake of convenience (see builder API and its supporting infrastructure).
- **Immutability:** Part of the initial idea was to make objects stored in `Shape` immutable. This was later changed to support use cases of various algorithms (namely transform and sweep).

So as it turns out, `Shape` didn't end up achieving its core goals. But it does bring with it a lot of complexity. So far, I've considered this just a cost of doing business, but I recently realized that it should be possible to remove `Shape` completely, which would be a radical simplification of the kernel data structures.

I'm thinking about the following measures:

- There's already a firm plan in place for validation. See the aforementioned #696.
- Object identity is simply not important right now, and it can be ignored.
- Immutability can be achieved in simpler ways, using standard Rust practices. The algorithms that rely on object mutability can probably be updated not to. If I recall correctly, transform only needs that because the complexity brought by `Shape` made its implementation really complicated, and introducing mutability was a simplification. For sweep, I have various vague ideas that I still need to explore.

My idea for the `Shape`-less future is to simply keep the object types as they are right now, roughly, with two important differences:

- Make all their fields private, and only allow access to them via accessor methods that return shared references.
- Instead of referring to other objects via `Handle`s, just have each objects own a copy of the referred to object.
  - Any potential for inconsistency can be taken care of by the validation infrastructure.
  - The redundancy this introduces is likely not critical right now, as there's a lot of redundancy already between different instances of `Shape`. I don't think things will get any worse, performance-wise.
  - If the redundancy ever turns into a problem (for example, though too-high memory usage), maybe some simpler form of central data structure can be re-introduced later.
- Maybe some supporting infrastructure will be required to manipulate objects efficiently and conveniently. That is yet to be determined.

So that's basically it. Many details are yet to be figured out, but this is the rough direction that I'd like to go in.

This issue isn't blocked right now, as all of the areas it covers can be worked on or experimented with immediately, but its completion is blocked on #696.

[...]

Issue #568, opened 2022-05-11
Approximate faces in surface coordinates


hannobraun:
When approximating a face, the points that make up the approximation are generated as 3D points. This is a bit nonsensical, as approximation could -theoretically- just as well be done in 2D surface coordinates, and those 3D points are immediately converted to 2D coordinates anyway, for the triangulation.

This has been fine for a while, but now I discovered a bug in the 3D to 2D conversion logic: https://github.com/hannobraun/Fornjot/issues/567

And in fact, converting 3D approximations into 2D seems to be the only use case for that broken code.

As best I can tell, #567 would be non-trivial to fix and might cause a loss in generality (`SweptCurve` would likely need to be replaced with separate structs for planes and cylinders, which would negatively affect our ability to support more curve types in the future). Therefore, a different solution could be to just remove the need for that code existing in the first place, by creating approximations in 2D, avoiding the detour through 3D.

This would also have positive effects on #250, which could become much easier to implement. On the negative side, this would add some complexity to the representations of edges and faces, which would have to keep track of their bounding vertices/edges in their local coordinate system.

I'll try to go ahead and implement this anyway, as I believe it would result in cleaner code overall, in addition to solving the immediate problem (#567). There might be some problems that I'm not seeing right now though. We'll see.

hannobraun:
I've made some good progress towards achieving this, but getting this over the finish line will take some more effort. The problem is that the next step is going to require careful modification of quite a lot of code, which is not only tedious, but error-prone. And most of that code shouldn't exist in the first place. Its existence reflects a lack of suitable methods of the `Shape` API.

#399 is probably a partial solution to this problem, but there are other new APIs that could help here (I've left some thoughts in the commits in #544).

I will look into this now, as solving this issue (and replacing all that problematic code) should make further progress here much easier.

hannobraun:
I've removed all the blockers, and am back to working on this issue directly.

hannobraun:
I have more progress towards this issue in local branches, but there's a bug in there that I haven't tracked down yet. I'm now working on #613, which should make the source of that bug quite obvious.

hannobraun:
#613 is taken care of (for now), and I'm back to working on this issue directly.

J-F-Liu:
If the purpose of 3D to 2D conversion is to do triangulation, is it possible and better to do triangulation on approximated 3D points without converting back to 2D surface coordinates?

hannobraun:
[...]

PR #723, opened 2022-06-23
Rewrite sweep algorithm

hannobraun:
The sweep algorithm has been a pain in my ass for a long time. It seems like it should be easy, but getting the details right has always been challenging, and the algorithm has turned into a real mess. (And despite being a huge mess, I believe it *still* doesn't get the details right. I'm now sure that a lot of bugs are hidden in there, that would be exposed as we get closer to #568.) I've attempted to simplify it a bunch of times, and some things about it have gotten better, but it never quite worked out as I hoped.

In addition to the aforementioned problems, there's another issue with it: it uses a lot of the problematic features of `Shape`, and as long as it does, removing `Shape` outright won't be possible (#697). Since I *really* want to do that (it will make so many things easier going forward), fixing the sweep algorithm has become a priority.

I believe the time for half-measures is over. It's time to get rid of this cadaver, and just rewrite the whole thing. This is what I've started doing in this pull request. First results are encouraging. The new version of the algorithm correctly generates the side faces, which has traditionally been the primary problem area. I believe getting the bottom and top faces working is just a matter of putting the effort in.

### DIFF
--- a/crates/fj-kernel/src/algorithms/mod.rs
+++ b/crates/fj-kernel/src/algorithms/mod.rs
@@ -12,7 +12,7 @@ pub mod intersection;
 
 pub use self::{
     approx::{CycleApprox, FaceApprox, InvalidTolerance, Tolerance},
-    sweep::sweep_shape,
+    sweep::sweep,
     transform::transform_shape,
     triangulation::triangulate,
 };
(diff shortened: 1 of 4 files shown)


hannobraun:
I've finished rewriting the sweep algorithm. I'm now working on cleaning up the branch and getting it ready for merge.

hannobraun:
Cleaned up the branch. Added more explanation to the commit messages. Ready for merge.

Issue #741, opened 2022-06-29
`export-validator` should export files to temporary directory


hannobraun:
[`export-validator`](https://github.com/hannobraun/Fornjot/tree/main/tools/export-validator) is a tool that exports all models from the [`models/`](https://github.com/hannobraun/Fornjot/tree/main/models) directory to 3MF, and checks that those exported models are valid. This is an important tool for detecting bugs, and it is included in the CI build, and the local `just build` command.

`export-validator` places the exported files in the repository root, which is annoying. It should create a temporary directory instead, and just place the files there.

I'm labeling this as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as the changes required are limited to one tool, and no knowledge of Fornjot as a whole is required. I've successfully used the [`tempfile`](https://crates.io/crates/tempfile) for this kind of thing in the past.

hannobraun:
This has been addressed by @eLVas in #749. Thank you!

PR #749, opened 2022-06-29
Use temporary directory for model export

eLVas:
Resolve issue #741


hannobraun (approved):
Thank you, @eLVas! Looks and works great!

Not sure what's up with CI right now. Seems to be stuck (on [another pull request](https://github.com/hannobraun/Fornjot/pull/748) too). I'll observe for a while and maybe try to re-trigger it, if it doesn't resolve itself.

PR #758, opened 2022-07-01
Only calculate focus point when needed

jeevcat:
Currently, on every type of input event (include mouse movement), the focus point is calculated. This can be an expensive operation for complex meshes.

If the mouse is moved just right (try small circular motions), many mouse movement events will be triggered, each causing a focus point recalculation. All of these mouse movement events must be handled before a new frame is rendered.

This PR just moves focus point calculation into the mouse click events which need it, preventing the much more frequent mouse movement events from triggering the calculations.

### DIFF
--- a/crates/fj-viewer/src/input/handler.rs
+++ b/crates/fj-viewer/src/input/handler.rs
@@ -4,7 +4,7 @@ use fj_interop::mesh::Mesh;
 use fj_math::{Point, Transform, Vector};
 
 use crate::{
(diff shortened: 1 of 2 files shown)


hannobraun:
Thank you, @jeevcat! I'll take a closer look later, but this seems like a nice improvement.

hannobraun (approved):
Just tested this. Works as great as it looks. Thank you, @jeevcat!

PR #763, opened 2022-07-01
Updated `egui` + `wgpu` + `winit` proof-of-concept integration

follower:
Here's the updated `egui` + `wgpu` + `winit` proof-of-concept integration for Fornjot, since I figured out & fixed what was causing the issue I ran into when I was working on it a month ago.

GH seems to think it's merge-able with current `main`.

Feel free to use/modify as you see fit--no idea when I'll be through next but I did at least want to follow through on my intention this far & hope it's helpful to Fornjot's progress in some way.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,6 +84,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "arboard"
+version = "2.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dc120354d1b5ec6d7aaf4876b602def75595937b5e15d356eb554ab5177e08bb"
+dependencies = [
+ "clipboard-win",
+ "log",
+ "objc",
+ "objc-foundation",
+ "objc_id",
+ "parking_lot 0.12.1",
+ "thiserror",
+ "winapi",
+ "x11rb",
+]
+
+[[package]]
 name = "arrayvec"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


hannobraun:
Thank you, @follower, this is great work! I really appreciate that you followed through and didn't leave me hanging with a tantalizing but unmergable pull request :smile: 

I've taken a quick look at the code, and aside from the minor Clippy/rustfmt stuff, the only real problem I see is that it re-adds a dependency on winit to `fj-viewer`. `fj-window` was extracted out of `fj-viewer` with the specific intent to break that dependency, and enable use of `fj-viewer` with other windowing solutions. To be clear, this is not a criticism of your work. I understand that this is a proof of concept, and such a limitation is perfectly acceptable. I'm just explaining the context to you, or anyone else who comes across this pull request and might be interested.

I'm actually planning to publish a new Fornjot release, and I'll start looking into that on Monday (might take me a bit to actually do it; there are some problems with the release automation). Right after that, I want to get this merged. My first impression is, that it shouldn't be too hard to reshuffle things a bit. I might either just do that, or merge it as-is and open an issue to track the unwanted dependency. Either way, we'll get a GUI that's much more capable than what we have now, and can iterate on that further.

Until this pull request is merged, I will hold off on merging any other pull requests that could cause conflicts with it.

hannobraun:
Okay, I've squashed the commits to make everything a bit more manageable, and made some small (and automated) changes to the bigger commit that adds all the code, as described in the commit message.

CI should pass now, and I'm happy with merging it in this state. I'll open follow-up issues once I have merged it.

Thanks again, @follower! This should provide a nice base for future improvements.

Issue #678, opened 2022-06-10
Moving the model should work without a focus point


hannobraun:
Fornjot's input/graphics code has the concept of a focus point: the point where the mouse cursor touches the model. This focus point is used to process input: If you rotate the model (move the mouse while pressing the left mouse button) while the mouse hovers over it, the model will be rotated around the focus point. When moving the model (move the mouse while pressing the right mouse button), the model is moved such, that the same point always stays under the mouse cursor, making the movement feel very natural.

Unfortunately, moving the model *only* works, if the mouse is touching the model (i.e. if there is a focus point). This can be quite confusing, if you want to move the model, but aren't currently touching it with the mouse (either by accident, or because you don't know that you need to touch it).

Moving the model should always work. If there is no focus point, a fallback should be used. This has already been done for rotation, and it would make sense to make sure that both moving and rotation use the same mechanism to compute the fallback point (making sure they are always in sync). I believe that rotation is currently using the coordinate system origin (0/0/0) as the fallback point, which is not ideal. I think it would be better, to use the center point of the models bounding volume.

All input code lives in `fj-viewer`. The movement code lives in [`movement.rs`](https://github.com/hannobraun/Fornjot/blob/74183a2e14ce63323cee02c5cd9f6bc4780373ff/crates/fj-viewer/src/input/movement.rs). The rotation code lives in [`rotation.rs`](https://github.com/hannobraun/Fornjot/blob/74183a2e14ce63323cee02c5cd9f6bc4780373ff/crates/fj-viewer/src/input/rotation.rs). The `FocusPoint` struct is [defined in `camera.rs`](https://github.com/hannobraun/Fornjot/blob/74183a2e14ce63323cee02c5cd9f6bc4780373ff/crates/fj-viewer/src/camera.rs#L224-L237). Maybe it makes sense to just replace the `FocusPoint` struct with a regular `Point<3>`, and set it to the fallback value, if necessary, right [where it is computed in `camera.rs`](https://github.com/hannobraun/Fornjot/blob/74183a2e14ce63323cee02c5cd9f6bc4780373ff/crates/fj-viewer/src/camera.rs#L129-L164).

I'm labeling this as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this issue is focused on input code only, and doesn't require much specific knowledge of Fornjot.

PR #764, opened 2022-07-03
Simplify zoom and adapt for larger models

### LINKED ISSUE #25: Zooming is way too slow for large models
Zooming is way too slow with large models. With the `enclosure` model, zooming basically does nothing.

Right now, there's a given standard zoom speed that is limited by how close the model is, to make viewing details up-close practical. I think this should probably be changed to make zoom speed dependent on distance to the model. Basically, however far away you are, zoom speed should work comfortably, never being too fast or too slow.

If anyone wants to look into this, basically all the relevant code is in [`zoom.rs`](https://github.com/hannobraun/fornjot/blob/cdd7e6e099bce9ff377a28f2ab959bef3c2292e5/src/input/zoom.rs). Specifically the [speed limiting code](https://github.com/hannobraun/fornjot/blob/cdd7e6e099bce9ff377a28f2ab959bef3c2292e5/src/input/zoom.rs#L111-L117) is interesting, and would have to be replaced when fixing this.

### LINKED ISSUE #677: Improve zooming behavior
The zooming behavior is a bit too finicky. It is quite easy to accidentally reverse the direction when trying to stop the zoom, for example.

The way zooming works, roughly, is that it checks how many scroll events (coming from the mouse wheel, for example) were generated over a recent time frame, and computes a zoom speed based on that. If it detects a scroll event that goes into the opposite direction as the current zoom speed, it will stop the zoom.

The design intent here is to provide appropriate control over the zooming behavior over a wide range of situations:
- You are zoomed in closely, but want to zoom out quickly to get an overview? Turn the mouse wheel quickly to generate a fast zoom.
- You are inspecting some detail of the model, and need to control the zoom very precisely? Zoom slowly by turning the mouse wheel slowly, and stop that zoom exactly where you want, by given the mouse wheel a turn in the opposite direction.

I think this design intent can be achieved using the current approach, but it needs some tuning to feel good. All input is handled in `fj-viewer`, and zooming specifically is done in [`zoom.rs`](https://github.com/hannobraun/Fornjot/blob/74183a2e14ce63323cee02c5cd9f6bc4780373ff/crates/fj-viewer/src/input/zoom.rs). Using a completely different approach that realizes the same design intent would also be a perfectly acceptable resolution to this issue.

Please also note that there is a related issue (#25) that is specifically about zooming speed. It might make sense to tackle both issues concurrently.

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as the code that governs zooming lives in a single file, making this a pretty focused issue that doesn't require any knowledge of Fornjot specifics.

jeevcat:
This PR removes the interpolated zoom system and replaces it with a basic instantaneous zoom system. The new algorithm moves the camera proportionally based on the current distance from the focus point. This means smaller changes in zoom when close to an object and larger changes in zoom when further away.

Please treat this PR as more of a proposal, as I have I deviated significantly from the initial design.

Close #677.
Close #25.

### DIFF
--- a/crates/fj-viewer/src/input/event.rs
+++ b/crates/fj-viewer/src/input/event.rs
@@ -9,7 +9,17 @@ pub enum Event {
     Key(Key, KeyState),
 
     /// The user scrolled
-    Scroll(f64),
+    Scroll(MouseScrollDelta),
+}
+
+/// Describes a difference in the mouse scroll wheel state.
+pub enum MouseScrollDelta {
+    /// Amount in lines to scroll in the horizontal direction.
+    ///
+    /// Positive values indicate movement forward (away from the user).
+    Line(f64),
+    /// Amount in pixels to scroll in the vertical direction.
+    Pixel(f64),
 }
 
 /// A keyboard or mouse key
--- a/crates/fj-viewer/src/input/handler.rs
+++ b/crates/fj-viewer/src/input/handler.rs
@@ -1,5 +1,3 @@
-use std::time::Instant;
-
 use fj_interop::mesh::Mesh;
 use fj_math::Point;
 
@@ -24,24 +22,6 @@ pub struct Handler {
 }
 
 impl Handler {
-    /// Returns a new Handler.
-    ///
-    /// # Examples
-    /// ```rust no_run
-    /// // Store initialization time for camera zoom calculations
-    /// let instant = std::time::Instant::now();
-    /// let input_handler = fj_viewer::input::Handler::new(instant);
-    /// ```
-    pub fn new(now: Instant) -> Self {
-        Self {
-            cursor: None,
-
-            movement: Movement::new(),
-            rotation: Rotation::new(),
-            zoom: Zoom::new(),
-        }
-    }
-
     /// Returns the state of the cursor position.
     pub fn cursor(&self) -> Option<Position> {
         self.cursor
@@ -52,7 +32,6 @@ impl Handler {
         &mut self,
         event: Event,
         screen_size: Size,
-        now: Instant,
         mesh: &Mesh<Point<3>>,
         camera: &mut Camera,
         actions: &mut Actions,
@@ -112,12 +91,24 @@ impl Handler {
     pub fn update(
         &mut self,
         delta_t: f64,
-        now: Instant,
         camera: &mut Camera,
-        size: Size,
+        screen_size: Size,
         mesh: &Mesh<Point<3>>,
     ) {
-        self.zoom.apply_to_camera(delta_t, camera);
+        let focus_point = camera.focus_point(screen_size, self.cursor(), mesh);
+        self.zoom.apply_to_camera(delta_t, focus_point, camera);
+    }
+}
+
+impl Default for Handler {
+    fn default() -> Self {
+        Self {
+            cursor: None,
+
+            movement: Movement::new(),
+            rotation: Rotation::new(),
+            zoom: Zoom::new(),
+        }
     }
 }
 
--- a/crates/fj-viewer/src/input/handler.rs
+++ b/crates/fj-viewer/src/input/handler.rs
@@ -1,16 +1,15 @@
 use std::time::Instant;
 
 use fj_interop::mesh::Mesh;
-use fj_math::{Point, Transform, Vector};
-
-use crate::{
-    camera::Camera,
-    screen::{Position, Size},
-};
+use fj_math::Point;
 
 use super::{
     event::KeyState, movement::Movement, rotation::Rotation, zoom::Zoom, Event,
     Key,
+};
+use crate::{
+    camera::Camera,
+    screen::{Position, Size},
 };
 
 /// Input handling abstraction
@@ -39,7 +38,7 @@ impl Handler {
 
             movement: Movement::new(),
             rotation: Rotation::new(),
-            zoom: Zoom::new(now),
+            zoom: Zoom::new(),
         }
     }
 
@@ -102,7 +101,7 @@ impl Handler {
             }
 
             Event::Scroll(delta) => {
-                self.zoom.push_input_delta(delta, now);
+                self.zoom.push(delta);
             }
 
             _ => {}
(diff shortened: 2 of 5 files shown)


hannobraun:
Thank you for the pull request, @jeevcat!

I haven't had a chance to take a close look and test this yet. I will do so later. If it works reasonably well, I will likely merge it. The current system isn't working so well, after all, and this is much less code.

jeevcat:
I was just reading the description of #678, and noticed that I'm also using the origin as a fallback for the focus point. If you're happy with the new zoom behavior you could just merge this PR and I'll make the suggest changes (always fallback to center of bounding box of model) in a following PR addressing #678. Otherwise, I could extend this PR with the changes.

jeevcat:
I've tested again and 2.0 is fine for my PC with a mouse wheel but is maybe a tiny bit too sensitive for my MacBook's trackpad. 1.0 seems pretty reasonable for me on both though. What do you think?

hannobraun:
> I've tested again and 2.0 is fine for my PC with a mouse wheel but is maybe a tiny bit too sensitive for my MacBook's trackpad. 1.0 seems pretty reasonable for me on both though. What do you think?

`1.0` seems a tiny bit too slow with the mouse wheel, but I would accept it :smile: 

After some more experimentation, `2.0` seems a bit too fast though. I've settled on `1.5` as my new favorite.

Regarding the trackpad, please check out this piece of code in `fj-window`:
https://github.com/hannobraun/Fornjot/blob/db910c73dc0e546dd6fda17747a6546d7cae0084/crates/fj-window/src/run.rs#L169-L176

One of those variants provides the value for mouse wheel, the other for trackpad (I forget which is which; this could use some comments). Maybe you could play around with that to find a value that works for you, then I can double-check against my own laptop?

jeevcat:
@hannobraun I've moved the zoom speed decision down into zoom.rs so we could remove the magic number in run.rs. Let me know how the behavior feels now.

crates/fj-viewer/src/input/event.rs
```diff
@@ -9,7 +9,17 @@ pub enum Event {
...
+}
+
+/// Describes a difference in the mouse scroll wheel state.
+pub enum MouseScrollDelta {
+    /// Amount in lines to scroll in the horizontal direction.
+    ///
+    /// Positive values indicate movement forward (away from the user).
+    Line(f64),
+    /// Amount in pixels to scroll in the vertical direction.
+    Pixel(f64),
```
hannobraun, line 22: A few issues with the documentation here:
- It says `Line` is horizontal, but they're both vertical.
- I believe the sign has the same meaning for both variants, so it probably makes sense to move the comment about positive values to the struct-level doc comment.

hannobraun (commented):
Thanks again, @jeevcat! I'll test this later, once I get a chance.

I'm not sure that structuring the mouse scroll event like this is the best way to do it. A potential alternative Fornjot application that uses `fj-viewer` but not `fj-window` might need its own tuning values anyway, if whatever windowing solution it uses doesn't line up with winit in that regard. And even the winit-based `fj-window` might need different tuning values for different contexts (for example with touch-based input, but I'm just speculating).

In any case, this is fine for now. I'm happy to merge it like this, if it means zoom works better afterwards. I'm just speculating that we might want to revert that particular change at some point.

I also left a comment about some minor issues in the code.

hannobraun (approved):
Okay, I decided to make some time for testing right now. Which was a strategic decision about what should go into the [next release](https://github.com/hannobraun/Fornjot/pull/779), and not at all an obvious ploy to further procrastinate on writing the release announcement.

Feels great, both with a mouse and a trackpad! I'm going to merge this right now. Any further improvements to architecture and/or doc comments can be done later.

hannobraun:
> When you get the chance, let me know what you're thinking and I'll happily tidy things up.

Basically just what I said in [this review](https://github.com/hannobraun/Fornjot/pull/764#pullrequestreview-1029942284). The documentation I commented on is an obvious opportunity for improvement, but I don't know, if it makes sense to tackle the architectural stuff now (i.e. whether to revert your zoom event change or not).

Might be best to just wait and address this later, once we have more information. What do you think?

jeevcat:
> I'm not sure that structuring the mouse scroll event like this is the best way to do it. A potential alternative Fornjot application that uses `fj-viewer` but not `fj-window` might need its own tuning values anyway, if whatever windowing solution it uses doesn't line up with winit in that regard. And even the winit-based `fj-window` might need different tuning values for different contexts (for example with touch-based input, but I'm just speculating).

So you're thinking move the tuning into fj-window, instead of where I put it in fj-viewer? My reasoning was that fj-viewer already exposes events like `CursorMoved`, `Key` and `Scroll`. If it exposed more abstract events like "Zoom" and "Pan" then something like `ZoomDelta` (which would be calculated in fj-window) instead of the current `MouseScrollDelta` would make more sense. But because we're passing these lower-level events, passing the type of scroll input made sense to me. What are your thoughts?

I'll fix up the documentation errors in a new PR now.

hannobraun:
I figure that tuning outside of `fj-viewer` might be necessary anyway, since different sources of input events might expose different values, and `fj-viewer` really can't (or shouldn't!) know about where these events are from. So yeah, moving the tuning outside might make sense. The event would then just transport "one unit of zoom", whatever that means, and `fj-window` would be responsible for making that feel good.


> My reasoning was that fj-viewer already exposes events like `CursorMoved`, `Key` and `Scroll`. If it exposed more abstract events like "Zoom" and "Pan" then something like `ZoomDelta` (which would be calculated in fj-window) instead of the current `MouseScrollDelta` would make more sense. But because we're passing these lower-level events, passing the type of scroll input made sense to me. What are your thoughts?

Yeah, that reasoning is perfectly valid. That code is probably not the best thing to base your reasoning on though, but how would you know that, if I didn't document my intent properly :smile: 

I think having those higher-level events you talk about makes total sense. Here's my latest thinking, based on this discussion and previous ones:

The reason those events exist, is because the winit-specific stuff was moved out of `fj-viewer`. The reason that was done, was to enable alternative applications that don't use winit (one contributor was working on something, but I haven't heard about that for a while). We shouldn't assume that those alternative applications are similar to the current `fj-app` (they might be optimized for touch interfaces, for example), so exposing higher-level events makes more sense than exposing the current lower-level ones.

hannobraun:
> Let me open a follow up PR which makes the input interface to fj-viewer more high-level and the we can discuss whether it makes sense to merge.

Sounds great! :+1:

PR #766, opened 2022-07-04
Check for docs warnings in CI

### LINKED ISSUE #271: Check documentation in CI build
The rustdoc documentation had accumulated a few warnings that I fixed (https://github.com/hannobraun/Fornjot/pull/270). This should be checked in the CI build. A check can be added to [`workflows/test.yml`](https://github.com/hannobraun/Fornjot/blob/main/.github/workflows/test.yml).

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this is a self-contained change that doesn't require knowledge of Fornjot.

jeevcat:
Close #271


crates/fj-kernel/src/validation/mod.rs
```diff
@@ -28,15 +28,14 @@
...
-};
-
 use std::{collections::HashSet, ops::Deref};
 
 use fj_math::Scalar;
 
+pub use self::{
+    coherence::{CoherenceIssues, CoherenceMismatch},
+    uniqueness::UniquenessIssues,
+};
```
hannobraun, line 38: Could you please revert this change? Placing those `pub use` statements right below the module definitions is quite deliberate, as they are closely related to them, and are part part of the module's public interface, which makes them distinct from the other `use` statements. I believe that mixing `use` and `pub use` makes it harder to follow what's going on.

hannobraun (changes requested):
Thank you, @jeevcat, this is awesome!

I've left one comment with a change request, but that's only a minor thing. Once that is addressed, we can merge this. The failing CI job isn't pretty, but better than not checking for those warnings at all.

> I fixed a few basic hyperlink warnings, but I wasn't sure about those regarding `Handle`s and `Shapes` (I think I read that `Shape` is now "abolished"?).

Yes, `Shape` is gone and `Handle` with it. I'll take a look at the remaining warnings later, to see if those references can be removed, or if anything should get updated.

---

As a note to myself, I need to open the following issues after merging this:

- [x] Address remaining doc warnings; make doc warning job a prerequisite to merging
- [ ] Add check for doc warnings to [justfile](https://github.com/hannobraun/Fornjot/blob/main/justfile) ([context](https://github.com/hannobraun/Fornjot/blob/main/CONTRIBUTING.md#making-improvements))

crates/fj-kernel/src/validation/mod.rs
jeevcat: Sorry, my bad. Seems I had rust-analyser configured to auto-format with some non-default options. Reverted.

hannobraun: Thanks! :+1: 

If there's any way to better support your setup and make things easier for you, I'd be happy to merge any repository-specific configuration that helps with that. We already have VS Code and rustfmt config, as well as some others.

hannobraun (approved):

hannobraun:
Thanks again, @jeevcat!

The various dependabot pull requests are battling for supremacy right now, to see which one gets the honor of being merged first. Once they've sorted themselves out (which shouldn't be more than a few minutes), I'll merge this one.

crates/fj-kernel/src/validation/mod.rs
jeevcat: Thanks, but it was definitely just a problem on my end. 👍

hannobraun:
Okay, everything sorted! I've opened #772 and #773.

Issue #773, opened 2022-07-04
Check doc warnings in `just build`


hannobraun:
As of #766, documentation warnings are checked as part of the CI build. The same change hasn't been made to [`justfile`](https://github.com/hannobraun/Fornjot/blob/main/justfile) ([`CONTRIBUTING.md` has some context on what that file is for](https://github.com/hannobraun/Fornjot/blob/main/CONTRIBUTING.md#making-improvements)).

The check should be added there, to keep a local run of `just build` as close as reasonable to the CI build.

jeevcat:
Should docs building warnings be treated as errors in this case too?

hannobraun:
I just noticed an annoying side effect of setting the environment variables in `justfile`: Cargo won't use the existing artifacts from a previous variable-less build, meaning a `just build` after a regular build will trigger a full rebuild. But Cargo also won't put these new artifacts into a separate directory within `target/` (like it would when you build with a different `--target`), so the next regular build triggers a full rebuild again.

I wonder if there's a good way to work around this. Not failing and just showing warnings is probably not good enough, as the output from `just build` can get quite long, and it might be easy to miss warnings in there. Maybe it's possible to override Cargo's target directory for `just build`? Maybe to something like `target/just/`?

jeevcat:
Okay if we don't use `RUSTFLAGS` it seems to not trigger a full rebuild. e.g. try this out:

```
export RUSTDOCFLAGS := "-D warnings"

build:
    cargo clippy --all-features -- -D warnings
    cargo test --all-features
    cargo fmt --check
    cargo doc --no-deps --document-private-items --all-features --workspace
    cargo run -p export-validator
```

hannobraun:
Yes, but that could cause the problem I'm worried about: Warnings getting lost in the long output, without the build failing.

I just discovered `CARGO_TARGET_DIR`. Currently trying that out.

jeevcat:
Sorry, I didn't mention it explicitly: I added `-- -D warnings` to the `clippy` command above. This will still fail the build.

hannobraun:
> Sorry, I didn't mention it explicitly: I added `-- -D warnings` to the `clippy` command above. This will still fail the build.

Ah, got it. That seems fine then? At least I can't think of a problem.

---

The following is the result of the investigation I did since posting my previous comment here. It's probably moot now, but I figured I'd post it for the sake of completeness. Might come in handy in some way.

[...]

PR #780, opened 2022-07-06
Don't use RUSTFLAGS to deny clippy warnings

jeevcat:
This fixes a bug where cargo wouldn't use the existing artifacts from a previous variable-less build, meaning a `just build` after a regular build would trigger a full rebuild. 

See https://github.com/hannobraun/Fornjot/issues/773#issuecomment-1173957536


hannobraun (approved):
Thank you, @jeevcat!

PR #782, opened 2022-07-06
Verify that Cargo.lock contains all changes it should

### LINKED ISSUE #656: CI build should check whether `Cargo.lock` is up-to-date
Various changes, like adding/removing/updating dependencies, or adding/removing/updates crates in the workspace, results in automatic updates to `Cargo.lock`. It is possible to then accidentally not commit these updates alongside the change that caused them. If such a partial commit is merged into `main`, the next person to run `cargo build`/`cargo test`/... on `main`, will have their local copy of `Cargo.lock` updated.

This can be inconvenient, confusing, and can get in the way of working with Git history (like running `git bisect`). It would be great, if the CI build would fail, if any of it resulted in changes to `Cargo.lock`. I don't know, if there's an easy way to achieve this. If not, adding a custom tool to `tools/` and calling that from one of the CI jobs (maybe *Clippy*) would be an option.

jeevcat:
Closes #656

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,9 @@ jobs:
         with:
           command: build
           args: --all-features
+      # Verify that Cargo.lock contains all changes it should
+      - name: Check up-to-date Cargo.lock
+        run: git diff --exit-code Cargo.lock
       - name: Run `cargo test`
         uses: actions-rs/cargo@4ff6ec2846f6e7217c1a9b0b503506665f134c4b
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,9 +141,9 @@ jobs:
         with:
           command: build
           args: --all-features
-      # Verify that Cargo.lock contains all changes it should
-      - name: Check up-to-date Cargo.lock
-        run: git diff --exit-code Cargo.lock
+      # The repo shouldn't have any unstaged changes after a build
+      - name: Check up-to-date repo
+        run: git diff --exit-code
       - name: Run `cargo test`
         uses: actions-rs/cargo@4ff6ec2846f6e7217c1a9b0b503506665f134c4b
         with:


jeevcat:
The new task passes right now. Might make sense to open a dummy PR to test the non-passing case?

hannobraun (commented):
Thank you, @jeevcat, that's a neat way to do it!

Two thoughts:
- Should we expand the check to all files, not just `Cargo.lock`? That might prevent other kinds of accidents, but I don't know specifically which.
- The same check should go into `justfile`.

> The new task passes right now. Might make sense to open a dummy PR to test the non-passing case?

Maybe just push a non-working commit to this pull request, then remove it before merging? That way we could test it right here.

Adding a dependency to any `Cargo.toml`, without committing the changes to `Cargo.lock`, should be enough to trigger the new check.

jeevcat:
Okay the commit above checks all files, not just `Cargo.lock`.

Next I'll push a commit which should fail the CI.

hannobraun:
Nice, it fails!

jeevcat:
The commit failed successfully 😄 I'll revert this commit now.

> The same check should go into `justfile`.

Not sure exactly how we should implement this. What is the use case? We don't currently have control over when `just build` is run, so there would be a lot of false positives, depending on what the state of the local repo is.

hannobraun (approved):
Nice! Thanks again, @jeevcat! Merging.

> Not sure exactly how we should implement this. What is the use case? We don't currently have control over when `just build` is run, so there would be a lot of false positives, depending on what the state of the local repo is.

The use case for `just build` is to allow developers to figure out whether their changes would pass CI, without the need to actually run CI (which might take longer than a local build, or they might not be ready to open a pull request yet). As such, `just build` should be as close to the CI build as practical.

But this isn't very critical in this case, and you are right that this might cause false positives. Ideally, there would be some output from `just build` that explains why this check could fail and what to do about it. If you want to look into that, I'd be happy to merge it. If not, just as well. I'm sure you can find something more useful to do (in general, I mean; if that more useful thing involves working on Fornjot, all the better :smile:).

jeevcat:
A popular tool I've used previously is [pre-commit](https://pre-commit.com/). This might prove useful here. This will then remove the unknown of "when" the checks run. The only problem is long build times every time you want to commit.

hannobraun:
> The only problem is long build times every time you want to commit.

I think this would be prohibitive. Having to wait for a bit while preparing for submitting a pull request is one thing, but a full build on every commit? Rust's compile times are already bad enough...

Issue #71, opened 2022-01-21
Switch model system to WASM


hannobraun:
So far, we've been using dynamically linked libraries to load models into the host application. That works well enough for desktop platforms, but to run Fornjot on the web, with arbitrary models, it is necessary to compile models to WebAssembly and load that into the host application.

If at all possible, WebAssembly should be used for desktop platforms too, so we don't have to maintain two parallel systems. That has the additional advantage of providing sandboxing of models, improving security (the current system is basically unsound by design). Something like [Wasmtime](https://wasmtime.dev/) or [Wasmer](https://wasmer.io/) could be used, to embed WebAssembly support into the host application for desktop platforms.

Michael-F-Bryan:
Hi @hannobraun :wave: 

Something that has gone really well at work is using [wit-bindgen](https://github.com/bytecodealliance/wit-bindgen) for declaring the interface between the host and WebAssembly plugin.

The idea is you write a `*.wit` file containing all the functions and interface objects your host will provide to the guest (e.g. [`runtime-v1.wit`](https://github.com/hotg-ai/wit-files/blob/0c8769b2b43e84e17685a350b6ad2cd8ccbfadcf/rune/runtime-v1.wit)), plus a `*.wit` file that defines any functions the plugin will expose to the host (e.g. [`rune-v1.wit`](https://github.com/hotg-ai/wit-files/blob/0c8769b2b43e84e17685a350b6ad2cd8ccbfadcf/rune/rune-v1.wit)). From there you can use the `wit-bindgen` CLI to or some procedural macros to generate glue code for both host and guest. If you are familiar with Protocol Buffers, it's a very similar workflow.

The end result is you just need to implement a trait or two and browse your API docs to see what functions are available. We've already got hosts for loading WebAssembly plugins [in the browser](https://github.com/hotg-ai/wit-files/tree/main/rune/js) or [using `wasmtime`](https://github.com/hotg-ai/proc-blocks/blob/08c82fcbbc80c8da24674add9402f4f1bae92cab/xtask/src/metadata.rs#L45-L83) from a desktop.

One thing to be aware of is that (by design) WebAssembly can't reference memory from the host, so any data you give plugins access to will need to be copied in. This could get expensive if you are giving plugins direct access to meshes. One solution would be to provide high-level objects that do the expensive manipulation for the plugin.

hannobraun:
Hey @Michael-F-Bryan, great to see you around here!

[...]

PR #783, opened 2022-07-06
[POC] Refresh on (extracted) parameter update via UI.

follower:
Screenshot from 2022-07-07 02-36-14

Screenshot from 2022-07-07 02-36-38

Screenshot from 2022-07-07 02-36-48

[TODO: Add some context]


hannobraun:
Okay, I've now had a chance to look at this more closely, and check out how it works.

I think from a UI perspective, this is a perfect starting point! However, the approach used to load the default parameters from the model is the wrong one, in my opinion. If I understand correctly, it parses the source file of the model. This duplicates the parsing code that already exists in `fj-proc`, and ties the implementation to Rust as a modeling language, making it more difficult to add support for other languages via WASM (#71) or other means.

Now that we have `#[fj::model]`, we can handle parameters in a more robust way: By having `#[fj::model]` generate an API in the model that the host can query to learn which parameters exist, and what their values are. I would accept pull requests that take the code base in that direction, but I'd also like to note that I think it makes more sense to address #71 first, as any infrastructure between host and model would need to be updated anyway when making the switch to WASM.

All of this shouldn't be seen as a complaint about your work, @follower! The pull request is clearly marked as draft, and has "proof of concept" right in the title, so I certainly didn't expect you to address long-standing architectural issues. I just want to put this pull request into a wider context, and address how this work could go forward, in my opinion.

What do you think, @follower? Where do you see this pull request going from here?

hannobraun:
I've opened an issue with thoughts on what a host API that could properly support this functionality could look like: https://github.com/hannobraun/Fornjot/issues/804

hannobraun:
I've kept this pull request open, as I hoped that it could serve as a template for future work along those lines. However, due to the recent changes (see [A New Direction](https://www.fornjot.app/blog/a-new-direction/)), this feature is no longer in scope. Closing.

Thanks again for your work, @follower!

PR #806, opened 2022-07-12
Moving the model should work without a focus point

### LINKED ISSUE #678: Moving the model should work without a focus point

jeevcat:
- Fix performance regression when zooming
- Moving the model works without focus point
- Focus point now just falls back to center of AABB of model

Closes #678

### DIFF
--- a/crates/fj-viewer/Cargo.toml
+++ b/crates/fj-viewer/Cargo.toml
@@ -28,6 +28,10 @@ path = "../fj-interop"
 version = "0.7.0"
 path = "../fj-math"
 
+[dependencies.fj-operations]
+version = "0.7.0"
+path = "../fj-operations"
+
 [dependencies.egui]
 version = "0.18.1"
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -944,6 +944,7 @@ dependencies = [
  "egui-winit",
  "fj-interop",
  "fj-math",
+ "fj-operations",
  "raw-window-handle",
  "thiserror",
  "tracing",
--- a/crates/fj-viewer/src/camera.rs
+++ b/crates/fj-viewer/src/camera.rs
@@ -1,8 +1,8 @@
 //! Viewer camera module
 use std::f64::consts::FRAC_PI_2;
 
-use fj_interop::mesh::Mesh;
 use fj_math::{Aabb, Point, Scalar, Transform, Triangle, Vector};
+use fj_operations::shape_processor::ProcessedShape;
 
 use crate::screen::NormalizedPosition;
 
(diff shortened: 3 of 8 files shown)


crates/fj-viewer/Cargo.toml
```diff
@@ -28,6 +28,10 @@ path = "../fj-interop"
 version = "0.7.0"
 path = "../fj-math"
 
+[dependencies.fj-operations]
+version = "0.7.0"
+path = "../fj-operations"
+
```
hannobraun, line 34: I think it's a bit unfortunate to add the dependency on `fj-operations` and (indirectly) `fj-kernel`. It would probably be better to move `ProcessedShape` to `fj-interop` instead.

hannobraun (approved):
Thank you @jeevcat, this is great!

I've left one comment, but that's a minor thing, and I'm happy to merge this PR anyway.

crates/fj-viewer/Cargo.toml
jeevcat: Yeah I didn't really like adding the dependency either. Let me give your suggestion a go.

PR #809, opened 2022-07-13
Move ProcessedShape to fj-interop

jeevcat:
As suggested here: https://github.com/hannobraun/Fornjot/pull/806#discussion_r919934299

Unfortunately, I had to remove the docs link to `ShapeProcessor::process`, as it is not reachable as a dependency of the `fj-interop` crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -944,7 +944,6 @@ dependencies = [
  "egui-winit",
  "fj-interop",
  "fj-math",
- "fj-operations",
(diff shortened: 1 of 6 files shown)


hannobraun (approved):
Thank you, @jeevcat!

> Unfortunately, I had to remove the docs link to `ShapeProcessor::process`, as it is not reachable as a dependency of the `fj-interop` crate.

That's an acceptable loss, I think.

PR #814, opened 2022-07-14
Release operator: Wait for crate version to appear upstream

### LINKED ISSUE #538: `Release | Crates.io` in CD pipeline doesn't handle `cargo publish` errors
See this run, which has lots of errors: https://github.com/hannobraun/Fornjot/runs/6311042754?check_suite_focus=true#step:9:76

If a `cargo publish` fails, the whole step should fail. Maybe it would even be better to abort the step, if a single `cargo publish` fails.

`Crate::submit` in `release-operator` seems to be the relevant piece of code:
https://github.com/hannobraun/Fornjot/blob/6df520ae30c2202c381b7cde8489621d0e67d5f4/tools/release-operator/src/registry.rs#L152-L180

I'm not seeing any checking of the command's return value there, so adding that and failing the whole `release-operator`, if `cargo-publish` fails, would probably address this issue.

### LINKED ISSUE #788: Release automation can fail, because just-published crate is not found
Check out this run of the CD pipeline: https://github.com/hannobraun/Fornjot/runs/7235683927?check_suite_focus=true

`fj` is being published successfully, but right after, `fj-host` fails, because it can't find `fj`. I suspect this is due to caching on crates.io.

I'm not sure what the best way to fix this is, but here are some ideas:

- Wait for a short delay after publishing a crate.
- After publishing a crate, query crates.io until it is available.
- Retry after a short delay, if publishing a crate fails.
- Retry if `cargo publish` fails for this specific reason

hendrikmaus:
This change-set introduces a simple wait loop, with a timeout of 10 minutes, after the submission of each crate. Hence the operator will only move to the next crate in the list of crates to publish once the current one has fully populated.

Fixes #788

### DIFF
--- a/tools/release-operator/src/registry.rs
+++ b/tools/release-operator/src/registry.rs
@@ -53,6 +53,17 @@ impl Registry {
     }
 }
 
+#[derive(Debug)]
+struct CrateNotFoundError;
+
+impl std::fmt::Display for CrateNotFoundError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "crate was not found on crates.io")
+    }
+}
+
+impl std::error::Error for CrateNotFoundError {}
+
 impl Crate {
     fn validate(&self) -> anyhow::Result<()> {
         match self.path.exists() {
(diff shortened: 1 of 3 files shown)


tools/release-operator/src/registry.rs
```diff
@@ -53,6 +54,17 @@ impl Registry {
...
+struct CrateNotFoundError;
+
+impl std::fmt::Display for CrateNotFoundError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "crate was not found on crates.io")
+    }
+}
+
+impl std::error::Error for CrateNotFoundError {}
+
```
hannobraun, line 67: The same thing could achieve in a more compact way with `thiserror`, but this is fine too.

hannobraun (approved):
Thank you @hendrikmaus, this looks great!

Notes:
- Left a nitpick comment; not important.
- I think your PR references the wrong issue. This addresses #788, not #538.
- I'm leaving this PR open until #538 has been fixed, as per [my comment in #788](https://github.com/hannobraun/Fornjot/issues/788#issuecomment-1178740212).

hendrikmaus:
I guess you are right, sry for the confusion. I'll fix the reference and open another PR for the more pressing issue.

tools/release-operator/src/registry.rs
hendrikmaus: I did it like this, because I did not want to pull in another dependency for a single error. 

However, I will check if the crate is already part of the dependency tree because of something else. If so, we can just as well use it.

hannobraun: It's already used in a bunch of places. At least `fj-kernel`, and I think `fj-viewer` too.

In any case, not a big deal. I just wanted to mention the possibility.

hannobraun:
> I guess you are right, sry for the confusion. I'll fix the reference and open another PR for the more pressing issue.

Thanks!

tools/release-operator/src/registry.rs
hendrikmaus: Pulled it in 😇

hannobraun:
Thank you again, @hendrikmaus! I'll take another look on Monday.

hannobraun (dismissed):
Looks good to me. Thank you @hendrikmaus!\

If #834 fixes #538, I will merge this right after, before I re-trigger the automation again.

hannobraun:
#538 is fixed. My fix introduced a merge conflict though. Rebased. Will merge, as well as CI is green.

PR #832, opened 2022-07-17
Make using models more convenient

### LINKED ISSUE #831: Shape types should implement equality
As currently implemented, none of the geometric types in the `fj` crate implement `PartialEq`. Is there a specific reason for this, or was it done just as a forward-compatiblity measure? For example, because we might want to add trait/function-based geometry in the future which doesn't have a definition for value equality.

If possible, it'd be nice to `#[derive(PartialEq)]` because then model implementations can write unit tests to check they generate the correct shape.

As an example for when this could be useful, see the following [smoke test](https://github.com/Michael-F-Bryan/fornjot-plugins/blob/cd6b82e7a806965d12cf0fd1f1682eebb3055d3f/models/cuboid/tests/integration.rs#L5-L30) for my `cuboid` model.

```rust
#[test]
fn load_cuboid_and_generate_geometry() {
    let crate_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
    let target_dir = crate_dir.parent().unwrap().parent().unwrap().join("target");

    let model = Model::from_path(crate_dir.into(), Some(target_dir.into())).unwrap();

    let mut arguments = HashMap::new();
    arguments.insert("x".to_string(), "10".to_string());
    let actual = model.load_once(&Parameters(arguments)).unwrap();

    let rectangle = fj::Sketch::from_points(vec![
        [-5.0, -1.0],
        [10.0 / 2., -2.0 / 2.],
        [10.0 / 2., 2.0 / 2.],
        [-10.0 / 2., 2.0 / 2.],
    ])
    .with_color([100, 255, 0, 200]);
    let expected = fj::Sweep::from_path(rectangle.into(), [0., 0., 3.0]);

    // TODO: Enable this assertion when we can compare shapes for equality.
    // https://github.com/hannobraun/Fornjot/issues/831

    // assert_eq!(actual, expected);
    let _ = (actual, expected);
}
```

Michael-F-Bryan:
This adds `PartialEq` bounds to the geometry types in `fj` (fixes #831) and also makes it easier to populate a `Parameters` dictionary (calling `.to_string()` all the time can be annoying).

### DIFF
--- a/crates/fj-host/src/lib.rs
+++ b/crates/fj-host/src/lib.rs
@@ -295,9 +295,9 @@ impl Parameters {
     pub fn insert(
         &mut self,
         key: impl Into<String>,
-        value: impl Into<String>,
+        value: impl ToString,
     ) -> &mut Self {
-        self.0.insert(key.into(), value.into());
+        self.0.insert(key.into(), value.to_string());
         self
     }
 }
--- a/crates/fj-host/src/lib.rs
+++ b/crates/fj-host/src/lib.rs
@@ -21,6 +21,7 @@ use std::{
     collections::{HashMap, HashSet},
     ffi::OsStr,
     io,
+    ops::{Deref, DerefMut},
     path::{Path, PathBuf},
     process::Command,
     sync::mpsc,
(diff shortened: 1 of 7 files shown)


hannobraun:
Thank you, @Michael-F-Bryan! I'll take a look later.

hannobraun (dismissed):
Thank you, @Michael-F-Bryan, looks good!

Will rebase/merge when #828 is merged.

hannobraun:
Ah, now we got a conflict, and I'm running out of time for today. Feel free to rebase, @Michael-F-Bryan. If not, I can take a look tomorrow.

Michael-F-Bryan:
@hannobraun, I've rebased onto `main`.

hannobraun:
Thank you!

PR #841, opened 2022-07-18
Make model load errors more specific

Michael-F-Bryan:
This addresses the comments @hannobraun made when reviewing #828.


hannobraun (approved):
Thank you, @Michael-F-Bryan, looks great!

PR #843, opened 2022-07-20
feat: try to publish a crate until it succeeds

hendrikmaus:
_Fixes #788_ 

This change-set introduces a loop around the publish command for the crate in scope. It will repeatedly attempt to publish if the command fails. The loop is limited by a timeout as well as a maximum number of attempts.

### DIFF
--- a/tools/release-operator/src/registry.rs
+++ b/tools/release-operator/src/registry.rs
@@ -5,6 +5,7 @@ use std::fmt::{Display, Formatter};
 use std::path::PathBuf;
 use std::process::Command;
 use std::str::FromStr;
+use std::thread::sleep;
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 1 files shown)


tools/release-operator/src/registry.rs
```diff
@@ -168,30 +169,52 @@ impl Crate {
...
+
+            if dry_run {
+                command.arg("--dry-run");
+            }
+
+            let exit_status = command.status()?;
+
+            if !exit_status.success() {
+                match exit_status.code() {
+                    Some(code) => {
```
hendrikmaus, line 202: FYI: no need to check which code came back as there is only a single failure code https://doc.rust-lang.org/cargo/commands/cargo-publish.html#exit-status

PR #867, opened 2022-07-23
Add typechecking to model signatures

### LINKED ISSUE #849: `#[fj::model]` should check signature of model function
The `#[fj::model]` proc macro provides a convenient way for users to define Fornjot models. The attribute is applied to a function, and isolates the user from the low-level implementation details of the host/model interface.

Model functions need to return an `fj::Shape`, but right now, this compiles:

``` rust
#[fj::model]
pub fn model() {
    todo!()
}
```

There is still a compiler error, if the wrong type is used. This fails to compile:

``` rust
#[fj::model]
pub fn model() {
    "invalid return value"
}
```

And this is the error:

```
error[E0308]: mismatched types
 --> models/error/src/lib.rs:3:5
  |
1 | #[fj::model]
  | ------------ expected `Shape` because of return type
2 | pub fn model() {
3 |     "invalid return value"
  |     ^^^^^^^^^^^^^^^^^^^^^^ expected enum `Shape`, found `&str`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `error` due to previous error
```

Michael-F-Bryan:
This fixes #849 by placing the original model function inside our generated `extern "C" fn model()`, letting the Rust compiler make sure it type checks. From there, it's simple to call the original function with its parsed arguments and assert that the result is convertible to a `fj::Shape`.

I've also added some examples to the attribute's docs so we can make sure `#[model] fn model() { todo!() }` doesn't compile.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -929,6 +929,7 @@ dependencies = [
 name = "fj-proc"
 version = "0.8.0"
 dependencies = [
+ "fj",
  "proc-macro2",
  "quote",
  "serde",
--- a/crates/fj-proc/Cargo.toml
+++ b/crates/fj-proc/Cargo.toml
@@ -26,3 +26,6 @@ optional = true
 [dependencies.syn]
 version = "1.0.98"
 features = ["full", "extra-traits"]
+
+[dev-dependencies]
+fj = { path = "../fj" }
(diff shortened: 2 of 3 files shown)


hannobraun (approved):
Thank you, @Michael-F-Bryan, looks great!

PR #868, opened 2022-07-23
Embed the git commit in fj-app's version number

### LINKED ISSUE #850: Implement `fj-app --version`
`fj-app` should have a `--version` argument that returns the current version. Given that we release weekly, it would be great if this version wouldn't have to be updated manually. This should be straight-forward to achieve: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

While I would definitely merge a pull request that implements this straight-forward version, it would be great if there was some additional nuance:

- The ID of the Git commit that the binary was built from should be output as well.
- It should be clear whether the binary is from a published release or not.

I assume that both of those can be achieved by creating the version string in a build script, and passing it to the `fj-app` source code as an environment variable. The release automation could use an environment variable to signal the build script whether this is an official release build.

Michael-F-Bryan:
Fixes #850.

The version number for an official build will look something like this:

```console
$ FJ_OFFICIAL_RELEASE=1 cargo run --quiet -- --version
fj-app 0.8.0 (66de87d0)
```

Not having the `$FJ_OFFICIAL_RELEASE` variable set will add `unreleased` to the version number, and having a dirty git tree will add a `-modified` suffix to the commit hash.

```console
$ cargo run --quiet -- --version
fj-app 0.8.0 (4ef5109e-modified, unreleased)
```

When `git` doesn't start (i.e. because it's not installed), we silently leave out the commit hash. This is a bit of an edge case, but I've found it often happens when you are building an app inside a Docker container or a `cargo install` from crates.io because you typically don't have access to the repo's `.git/` folder.

We could use something like [`vergen`](https://docs.rs/vergen) or [`build_info`](https://docs.rs/build_info) to get more comprehensive output similar to `rustc --version --verbose`, but that felt like overkill and would have an impact on compile times.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -2,7 +2,7 @@ name: Continuous Deployment
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
 
 env:
   CARGO_TERM_COLOR: always
@@ -10,6 +10,10 @@ env:
   # Name of the crate from Cargo.toml
   # used to rename and upload the binaries
   PROJ_NAME: fj-app
+
+  # This lets our app know it's an "official" release. Otherwise we would get
+  # a version number like "fj-app 0.8.0 (8cb928bb, unreleased)"
+  FJ_OFFICIAL_RELEASE: 1
 
 defaults:
   run:
--- a/crates/fj-app/Cargo.toml
+++ b/crates/fj-app/Cargo.toml
@@ -10,7 +10,6 @@ repository = "https://github.com/hannobraun/fornjot"
 license = "0BSD"
 keywords = ["cad", "programmatic", "code-cad"]
 categories = ["encoding", "mathematics", "rendering"]
-build = "build.rs"
 
 
 [dependencies]
--- a/crates/fj-app/Cargo.toml
+++ b/crates/fj-app/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/hannobraun/fornjot"
 license = "0BSD"
 keywords = ["cad", "programmatic", "code-cad"]
 categories = ["encoding", "mathematics", "rendering"]
+build = "build.rs"
 
 
 [dependencies]
--- a/crates/fj-app/build.rs
+++ b/crates/fj-app/build.rs
@@ -63,11 +63,15 @@ fn git_description() -> Option<String> {
     println!("cargo:rerun-if-changed={}", head_file.display());
 
     if let Ok(contents) = std::fs::read_to_string(&head_file) {
-        let (_, branch) = contents
-            .split_once(' ')
-            .expect(".git/HEAD should point to a valid head file");
-        let commit_hash_file = project_root.join(".git").join(branch.trim());
-        println!("cargo:rerun-if-changed={}", commit_hash_file.display());
+        // Most of the time the HEAD file will be `ref: refs/heads/$branch`, but
+        // when it's a detached head we'll only get the commit hash and can skip
+        // the rerun-if-changed logic.
+
+        if let Some((_, branch)) = contents.split_once(' ') {
+            let commit_hash_file =
+                project_root.join(".git").join(branch.trim());
+            println!("cargo:rerun-if-changed={}", commit_hash_file.display());
+        }
     }
 
     Some(stdout.trim().to_string())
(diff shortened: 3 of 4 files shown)


crates/fj-app/Cargo.toml
```diff
@@ -10,6 +10,7 @@ repository = "https://github.com/hannobraun/fornjot"
 license = "0BSD"
 keywords = ["cad", "programmatic", "code-cad"]
 categories = ["encoding", "mathematics", "rendering"]
+build = "build.rs"
```
hannobraun, line 13: Should be unnecessary, no? At least I'm pretty sure I've never added this, beside working with build scripts before.

.github/workflows/cd.yml
```diff
@@ -11,6 +11,10 @@ env:
   # used to rename and upload the binaries
   PROJ_NAME: fj-app
 
+  # This lets our app know it's an "official" release. Otherwise we would get
+  # a version number like "fj-app 0.8.0 (8cb928bb, unreleased)"
+  FJ_OFFICIAL_RELEASE: 1
```
hannobraun, line 16: I believe this is incorrect. The CD workflow is run (and builds binaries for) every push to `main`. If a release is detected, those binaries are re-used for the actual release.

Note the `Operator | Deduce` step, which detects whether we're dealing with a release, or a regular push to `main`, and the `if: ${{ steps.release.outputs.release-detected == 'true' }}` bits that use that step's output.

Not sure how good your grasp on GitHub Actions is. We might need @hendrikmaus to help figure this out.

If this turns into a problem, we could consider removing the per-push-to-`main` build, and only build binaries for the release. I'm not sure how useful those per-push binaries are anymore, with weekly releases. Thoughts?

crates/fj-app/build.rs
```diff
@@ -0,0 +1,78 @@
...
+
+fn main() {
+    println!("cargo:rustc-env=FJ_VERSION_STRING={}", version_string());
+}
+
+fn version_string() -> String {
+    let pkg_version = std::env::var("CARGO_PKG_VERSION").unwrap();
+    let commit = git_description();
+
+    let official_release = std::env::var("FJ_OFFICIAL_RELEASE").is_ok();
```
hannobraun, line 14: Minor issue: If someone ever set `FJ_OFFICIAL_RELEASE=0`, this would be confusing.

hannobraun (changes requested):
Thank you, @Michael-F-Bryan! Looks good overall, but I found some potential issues. I left comments.

crates/fj-app/Cargo.toml
hannobraun: To be clear, this is just me wondering out loud. No change needs to be made.

.github/workflows/cd.yml
Michael-F-Bryan: Ahh, I just had a deeper look into how the release job works, and it seems like every push to `main` generates its own compiled binaries, but they only get promoted to a proper release if the "release operator" says so.

That makes things more complicated because `fj-app` won't know at compile time whether it is a proper release (the `release` job depends on `binaries`, but `Binaries | Compile` would need to know `steps.release.outputs.release-detected` - which is a dependency cycle).

crates/fj-app/build.rs
Michael-F-Bryan: Good point.

crates/fj-app/Cargo.toml
Michael-F-Bryan: Yeah, as of a couple years ago `cargo` will automatically run a `build.rs` if it finds one next to `Cargo.toml`. I've always put it in there to be explicit, but I guess it's not really necessary.

hannobraun (commented):
Thank you, @Michael-F-Bryan, the new commits look good!

The only remaining problem is that all CD binaries are blessed as release binaries, not just the actual release ones.

Ideas for addressing this:

1. Fix this problem by cleverly changing the CD workflow.
2. Fix this problem less cleverly: only build binaries for actual releases.
3. Ignore the problem, merge this PR, open an issue.

The more I'm thinking about this, the more I'm leaning towards solution 3. This is already a nice improvement, and there's no need to hold it back because of one edge case.

I'll think some more. I expect to make a decision later today.

hannobraun (approved):

hannobraun:
I've decided to just merge this pull request. I'll open a new issue about the wrongly labeled binaries.

hannobraun:
I've opened #883.

PR #885, opened 2022-07-28
Model driven api

### LINKED ISSUE #804: Switch to model-driven model system
Fornjot's current model system is what I've decided to call "host-driven". The host is in control and just calls into the model function, passing it everything it needs (i.e. the parameters). The model function then creates the model and returns it. The host then goes on to do with that whatever it needs to.

This works well enough for now, but I believe it won't hold up in the future:

- It's not great that every model needs to generate its own logic for handling parameters. Checking that the required ones are there, setting default values, and rejecting parameters that the model doesn't use could all be done on the host side. That would require some unnecessarily complicated API for querying model parameters.
- Same, if we want to support a GUI for setting model parameters.
- We will need a closer interaction between kernel and model to support more advanced modeling features, for example referencing an existing face to put a sketch on it.
- What if we want to support a mode in which the model is an application that uses Fornjot as a library? This would be useful to provide customizable models in a form that doesn't require end users to have Fornjot installed (on a website, for example). This would require a whole different API and a whole different way of writing models.
- Once we allow models to export different artifacts (like a sketch for export to DXF, in addition to a 3D model for export to 3MF), there needs to be an additional API to query which functions are there.

I believe that there is a better approach that can support all those use cases more simply:

- The model just has a simple `init`/`main` function without parameters.
- The only thing the host does out of its own initiative is to call this `init`/`main` function.
- The host provides an API that the model can call.
- Everything in this API would be `#[repr(C)]`/`extern "C"`, so it works with over the FFI boundary required by the dynamic library, but also to provide flexibility for future uses ([WASM](https://github.com/hannobraun/Fornjot/issues/71), other languages, etc.).
- In an alternative Fornjot-as-library scenario, the same API would be available to the model.
- The `init`/`main` function and some of the API calls (e.g. for loading parameters) are generated by the proc macro.
- The `fj` crate provides a convenient and idiomatic Rust wrapper, for the model code to call.

Credit goes to @Protryon, who has made me aware of those two different approaches and provided crucial input over in the [Matrix channel](https://matrix.to/#/#fornjot:braun-odw.eu).

# Implementation

Here are my thoughts on how an implementation of this issue could go:

1. Create a new crate (maybe `fj-host-api`) that defines (but doesn't implement) the host API. The whole crate would basically be an `extern "C" { ... }` block with a number of function definitions inside. See below for my thoughts on the specific API.
2. Depend on this crate in `fj-host` and implement the API there.
3. Depend on this crate in `fj`. Because the API is defined in a lightweight crate without the implementation, this shouldn't negatively affect the compile times of models.
4. Update `fj-proc` to generate the `init`/`main` function and calls to the API for loading parameters.
5. Update `fj-host` to call the new `init`/`main` function instead.
6. Update `fj-proc` to no longer generate the old and now unused model function.

Here's what I imagine the host API could look like:
``` rust
extern "C" {
    // Functions for loading model parameters. These would be  required for the
    // initial implementation.
    pub fn load_param_u32(name: *const c_char, default: u32) -> u32;
    pub fn load_param_f64(name: *const c_char, default: f64) -> f64;
    // Add other `load_param` functions for other types as required.

    // API for registering the 3D shape that the model exports. This could
    // probably be left for a later step. The initial implementation could keep
    // the return value from the `init`/`main` function.
    pub fn register_solid(solid: Shape3d);
}

// For the `register_solid` function, we'd need to move this here from the `fj`
// crate. `fj` could just re-export it.
#[repr(C)]
pub enum Shape3d {
    // ...
}
```

That's just an initial sketch, of course. I'm sure I'm missing something, and I'm sure there are better ways to do it.

Michael-F-Bryan:
This is an initial implementation of #804 based on the ideas I was trying in [Michael-F-Bryan/fornjot-plugins](https://github.com/Michael-F-Bryan/fornjot-plugins). The discussion in https://github.com/hannobraun/Fornjot/issues/804#issuecomment-1197704136 contains a more in-depth explanation of how it works.


hannobraun (changes requested):
Thank you, @Michael-F-Bryan, this is great! Looking forward to merging this!

I didn't go through this pull request line-by-line. This is a lot of code, and I figure it's more efficient to immerse myself in it over the next weeks and months of working with it, instead of trying to understand every detail right now. The CI build passes, so I'm confident it's working as intended, or close enough.

I did look at every commit though, and do have some comments:

- I think it's a bit unfortunate that within the `fj` crate, the code for defining models (everything that was there before this PR) is intermixed with the new code. I think it would be better to have the new code in a dedicated module, which is clearly documented as "don't worry about this, this is taken care of for you (by the proc macro)".
- If that were the case, then the `abi` module could be un-`#[doc(hidden)]`. I've worked with crates in the past that are hiding internal implementation details from the API reference, and I found this is often confusing, and just makes it more likely that I have to dive into the code.
- What's up with those `todo!`s that are still left in the code? Where those missed, or is that stuff that can be filled in later? (I mean, CI passes, so presumably the latter.)
- All that hand-written FFI code is unfortunate, but given that we'll most likely switch to WebAssembly and WIT in the future, it's fine. If that code turns into too much of a maintenance issue, we can accelerate the transition to WASM/WIT, by publishing our own fork of `wit-bindgen` to crates.io, as I suggested as a possibility in #804.

All of those comments are just my observations on where I see potential for improvement, or where I could see things go. None of them need to be taken care of in this pull request, and in fact I'd prefer if we could get this pull request merged quickly and take care of anything else later.

However, there is one thing that I see as a blocker: With this change, we're losing the convenient `#[fj::model]` syntax, and I'd rather keep that. That syntax is mentioned in the README and past release announcement, and we'd have to at least adapt the README to prevent false advertising.

I think it would be better to port the macro to the new infrastructure instead before merging this PR. What do you think?

Michael-F-Bryan:
@hannobraun can you have another look at this? I've updated the `#[fj::model]` macro to generate the new code and switched the models (`star`, `cubiod`, etc.) back to their original implementations.

> I think it's a bit unfortunate that within the fj crate, the code for defining models (everything that was there before this PR) is intermixed with the new code. I think it would be better to have the new code in a dedicated module, which is clearly documented as "don't worry about this, this is taken care of for you (by the proc macro)".

Agreed. The main barrier to splitting them is that we have a dependency cycle.

- We want to move the `register_model!()` macro to some hypothetical `fj-models` crate and re-export it under `fj`, however
- The `register_model!()` macro references `Host`, and
- the `Host` trait (via the `Model` trait) references `Shape`

> What's up with those todo!s that are still left in the code? Where those missed, or is that stuff that can be filled in later? (I mean, CI passes, so presumably the latter.)

Oops! I forgot to wire up converting the FFI-safe version of `Metadata` to the canonical representation. The host doesn't access metadata at the moment, so the `todo!()` was never hit.

> All that hand-written FFI code is unfortunate, but given that we'll most likely switch to WebAssembly and WIT in the future, it's fine. If that code turns into too much of a maintenance issue, we can accelerate the transition to WASM/WIT, by publishing our own fork of wit-bindgen to crates.io, as I suggested as a possibility in https://github.com/hannobraun/Fornjot/issues/804.

Yeah, it's not great. 

In theory, we could use `#[derive(StableAbi)]` derive from [the `stable_abi` crate](https://docs.rs/abi_stable/latest/abi_stable/) to generate stable versions of any type that gets passed across the FFI boundary. It also takes care of the vtables/trait object stuff I was writing by hand. It wasn't until *after* I wrote all that FFI code that I remembered the `stable_abi` crate :sweat_smile: 

Sorry for the delay, I spent most of last week out of the city and away from a computer.

hannobraun:
Thank you, @Michael-F-Bryan! I'll take another look later today.

hannobraun (approved):
Thanks again, @Michael-F-Bryan! Looks good!

I'm going to merge this PR now, and will address your comments in a separate comment.

hannobraun:
> > I think it's a bit unfortunate that within the fj crate, the code for defining models (everything that was there before this PR) is intermixed with the new code. I think it would be better to have the new code in a dedicated module, which is clearly documented as "don't worry about this, this is taken care of for you (by the proc macro)".
> 
> Agreed. The main barrier to splitting them is that we have a dependency cycle.
> 
> * We want to move the `register_model!()` macro to some hypothetical `fj-models` crate and re-export it under `fj`, however
> 
> * The `register_model!()` macro references `Host`, and
> 
> * the `Host` trait (via the `Model` trait) references `Shape`

But I was suggesting to collect the new code in a module within the `fj` crate. That should be possible, right?


> > All that hand-written FFI code is unfortunate, but given that we'll most likely switch to WebAssembly and WIT in the future, it's fine. If that code turns into too much of a maintenance issue, we can accelerate the transition to WASM/WIT, by publishing our own fork of wit-bindgen to crates.io, as I suggested as a possibility in #804.
> 
> Yeah, it's not great.
> 
> In theory, we could use `#[derive(StableAbi)]` derive from [the `stable_abi` crate](https://docs.rs/abi_stable/latest/abi_stable/) to generate stable versions of any type that gets passed across the FFI boundary. It also takes care of the vtables/trait object stuff I was writing by hand. It wasn't until _after_ I wrote all that FFI code that I remembered the `stable_abi` crate :sweat_smile:

I actually experimented with `stable_abi` a while ago, hoping I could use their `Vec`-equivalent. I remember hitting some problem, but don't recall the details.

In any case, the code exists now. If it turns into a problem before the WASM/WIT migration, we have options.


> Sorry for the delay, I spent most of last week out of the city and away from a computer.

No problem! Thank you for taking the time to update this PR and get it merge-ready!

Issue #33, opened 2022-01-10
Fornjot application can crash, if specific graphics features are not supported on target


hannobraun:
As of this writing, the graphics code uses one feature that might not be available everywhere: `wgpu::Features::POLYGON_MODE_LINE`

The code just requests that feature. If that features isn't available on the target (where the host application is executed), the host application would either fail with an error message, or even panic. This is not optimal, especially since `POLYGON_MODE_LINE` is only used for an optional drawing mode (rendering the lines of the triangle mesh).

Although I haven't seen or heard of any problems regarding this, it would be better to handle these optional features in a more robust way:
1. Check which of features we want to use are actually supported.
2. Only request those features that are actually available. Store that information somewhere.
3. Don't let the user enable anything that requires features which are not available.

Ideally, there would be a nice error message in the UI, to let the user know what the problem is. But for now, it would already be an improvement, if the host application didn't crash if such a feature wasn't available.

hannobraun:
I've marked this as `good first issue`, since it only requires some knowledge of the `wgpu` API, and very little knowledge of Fornjot.

The code that requests the features lives in `renderer.rs`:
https://github.com/hannobraun/fornjot/blob/64299e64c081dccd97e86b3d84a220cb2d4e3771/src/graphics/renderer.rs#L55-L62

This `wgpu` method can probably be used to determine which features are actually available:
https://docs.rs/wgpu/0.12.0/wgpu/struct.Adapter.html#method.features

I assume that which features are available would be stored in a field of `Renderer`. Probably makes sense to create a separate `features` module under `graphics`, with a `Features` struct to encapsulate this (I'm just musing here; don't feel bound by my ideas, if you think something else makes more sense).

Here's some drawing code that requires this feature to exist:
https://github.com/hannobraun/fornjot/blob/64299e64c081dccd97e86b3d84a220cb2d4e3771/src/graphics/renderer.rs#L197-L212

This code (and potentially other code like it) would need to check the features field of `Renderer` to determine whether it should execute. Eventually, this information should also be fed outside the renderer, so the UI knows what's up and can display a proper error message, but that could be done as an additional improvement down the line.

hannobraun:
This issue has been addressed in #902. Thanks again, @hekno25!

I'm closing this issue, as Fornjot can no longer crash for this reason. I've opened the more specific #907 as a follow-up.

PR #902, opened 2022-08-02
Store and check adapter features before using them

hekno25:
This commit solves #33 by adding a `wgpu::Features` field to the `Renderer` struct and checking it before drawing the mesh or debug lines. The renderer does not inform the user if it is unable to use the `POLYGON_MODE_LINE` feature but it should at least prevent crashing. If this is too rudimentary I'd be glad to keep working on it and make the gui prevent mesh or debug rendering when the feature is missing.

### DIFF
--- a/crates/fj-viewer/src/graphics/renderer.rs
+++ b/crates/fj-viewer/src/graphics/renderer.rs
@@ -44,6 +44,7 @@ impl std::fmt::Debug for EguiState {
 #[derive(Debug)]
 pub struct Renderer {
     surface: wgpu::Surface,
+    features: wgpu::Features,
     device: wgpu::Device,
     queue: wgpu::Queue,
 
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @hekno25, this looks great! Glad you found such a simple solution!

> The renderer does not inform the user if it is unable to use the `POLYGON_MODE_LINE` feature but it should at least prevent crashing. If this is too rudimentary I'd be glad to keep working on it and make the gui prevent mesh or debug rendering when the feature is missing.

Informing the user is definitely desirable, and if you want to work on that, that would be very welcome. But this is already an improvement in itself, and I'm glad to merge it.

In general, I prefer merging a series of small improvement, over a big pull request that does everything perfect. Small changes are easier to review, cause less merge conflicts (thus less work for the contributor), and they reduce the risk of wasted work.

PR #911, opened 2022-08-03
Add UI element that displays model compiling status

devanlooches:


crates/fj-host/src/lib.rs
```diff
@@ -260,22 +265,31 @@ impl Watcher {
     ///
     /// Returns `None`, if the model has not changed since the last time this
     /// method was called.
-    pub fn receive(&self) -> Option<fj::Shape> {
+    pub fn receive(&self, mut status: &mut StatusReport) -> Option<fj::Shape> {
+        status.update_status("Receiving updated shape...");
```
hannobraun, line 269: Hey @devanlooches, I think this line is definitely one of the problems. The `receive` method is called in `run`, at the beginning of every iteration of the event loop. Meaning whatever the status used to be, at the beginning of the event loop (and therefore before every render) it is overwritten by this.

This is not the only problem. Other status updates seem to have a similar effect, and I haven't had time to understand everything completely. But hopefully this is enough to get you unstuck, so you can make more progress.

If you hit any more problems, always feel free to ask!

devanlooches: Hi! Thanks for the help I'll see if I can figure out what I can do.

devanlooches: So I have the full error showing up finally! But since it is the full compiler error it is very badly formatted:
image
Do you have any thoughts on how to filter it?

devanlooches: Sounds good. I think there is a scroll area container that we can use to wrap around the label [see here](https://docs.rs/egui/0.18.1/egui/containers/struct.ScrollArea.html).

devanlooches: I will also update the docs for the StatusReport struct.

devanlooches: Something like this works:

```rust
egui::ScrollArea::vertical().max_height(16.0)
    .show(ui, |ui| ui.label("Test label\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\ntest2"));
```

hannobraun (changes requested):
Thank you, @devanlooches!

This looks almost good to merge now, but the CI build still fails, specifically the `cargo fmt` and `cargo clippy` steps.

You should be able to take care of formatting automatically by running `cargo fmt` and just committing the result. You can fix the Clippy error by running `cargo clippy` and following its suggestion.

To get a feel for whether the CI build will pass, you can run `just build` locally, as documented here: https://github.com/hannobraun/Fornjot/blob/main/CONTRIBUTING.md#making-improvements

crates/fj-host/src/lib.rs
hannobraun: Looks good! Looking forward to seeing that.

devanlooches:
For some reason the ci build passes. However locally when I ran just build it failed with a linker error on the export validator step.

hannobraun (approved):
Thank you, @devanlooches, looks good!

hannobraun:
Oh, I forgot to reply to this:

> However locally when I ran just build it failed with a linker error on the export validator step.

That's sounds like a bug! Could you open an issue with the full error message? Maybe it's something platform-specific, or I have something installed that's missing on your machine.

PR #913, opened 2022-08-03
Validate `Circle` on construction

hannobraun:


hannobraun:
Looks like this validation fails in spectacular fashion with the spacer and test models. Likely a pre-existing bug that is exposed here.

I'll take a look tomorrow. It's getting late.

hannobraun:
Turns out it it wasn't a pre-existing bug, just some stupidity in this pull request (the `Circle::b` method returned `self.a`). Everything builds now locally.

PR #919, opened 2022-08-05
Extend status report element

devanlooches:
Right now I added the full error inside of a stretchable and scrollable element on a right side bar.
What else would you like me to include in the status update element and also why did you want me to make it hold a vector?

### DIFF
--- a/crates/fj-interop/src/status_report.rs
+++ b/crates/fj-interop/src/status_report.rs
@@ -13,6 +13,9 @@ impl StatusReport {
 
     /// Update the status
     pub fn update_status(&mut self, status: &str) {
+        if self.status.len() >= 5 {
+            self.clear_status();
+        }
         self.status.push(status.to_string());
     }
 
@@ -22,7 +25,7 @@ impl StatusReport {
     }
 
     /// Reset status
-    pub fn clear_status(&mut self) {
+    fn clear_status(&mut self) {
         self.status.clear();
     }
 }
--- a/crates/fj-interop/src/status_report.rs
+++ b/crates/fj-interop/src/status_report.rs
@@ -2,25 +2,28 @@
 
 /// Struct to store and update status messages
 pub struct StatusReport {
-    status: String,
+    status: Vec<String>,
 }
 
 impl StatusReport {
     /// Create a new ``StatusReport`` instance with a blank status
     pub fn new() -> Self {
-        Self {
-            status: String::new(),
-        }
+        Self { status: Vec::new() }
     }
 
     /// Update the status
     pub fn update_status(&mut self, status: &str) {
-        self.status = status.to_string();
+        self.status.push(status.to_string());
     }
 
     /// Get current status
-    pub fn status(&self) -> &str {
-        self.status.as_str()
+    pub fn status(&self) -> String {
+        self.status.join("\n")
+    }
+
+    /// Reset status
+    pub fn clear_status(&mut self) {
+        self.status.clear();
     }
 }
 
--- a/crates/fj-host/src/lib.rs
+++ b/crates/fj-host/src/lib.rs
@@ -91,13 +91,25 @@ impl Model {
 
         let command = command_root
             .arg("build")
+            .arg("-q")
             .args(["--manifest-path", &manifest_path]);
         let exit_status = command.status()?;
 
         if exit_status.success() {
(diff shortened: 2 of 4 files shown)


hannobraun (commented):
Thank you, @devanlooches!

Do you think this could be placed into a corner (lower-right? doesn't matter, maybe), and made more broad than high? Then there would be fewer line breaks in the error message.

> What else would you like me to include in the status update element

No more ideas right now, but now that we have it, it's easy to add more when we need to :+1: 


> and also why did you want me to make it hold a vector?

I figured then we could display the last few status messages, and not just the last one.

devanlooches:
> Do you think this could be placed into a corner (lower-right? doesn't matter, maybe), and made more broad than high? Then there would be fewer line breaks in the error message.

We could place it there but the element is supposed to auto resize itself to the contents (I don't know why it breaks the error message). But it is also stretchable so you can extend it horizontally by dragging the edge outwards.

> I figured then we could display the last few status messages, and not just the last one.

How many status messages do you think should be kept before clearing it?

devanlooches:
I could make the element completely draggable that way the user can put wherever they want.

hannobraun (commented):
Thank you for the additional commits, @devanlooches, looks good!

> We could place it there but the element is supposed to auto resize itself to the contents (I don't know why it breaks the error message). But it is also stretchable so you can extend it horizontally by dragging the edge outwards.

Hmm, when I grab it at its edge, I can move it, but not resize it. After I stop pressing the left mouse button, it resets to its original position.

> I could make the element completely draggable that way the user can put wherever they want.

That might be ideal! Is that easy to do?

> How many status messages do you think should be kept before clearing it?

I don't know. Maybe 5?

It doesn't really matter. Once the mechanism is there to store/display multiple messages and limit them to a certain number, it's easy to adjust that number later, as we gain more experience with it.

devanlooches:
Cool I will implement all of this shortly its a pretty quick fix.

hannobraun:
Nice, thank you!

devanlooches:
Here is the fix. Unfortunately I'm having trouble putting the window on the bottom right by default. If anyone could help me with that it would be great.

devanlooches:
One other thing I was thinking is it might be a good idea to keep the 5 status messages unless there is a compiler error in which case it might be a good idea to clear them in order to save space for the compiler error message (And also not show it more than once).

hannobraun:
Thank you, @devanlooches, I'll take a look later!

crates/fj-interop/src/status_report.rs
```diff
@@ -2,25 +2,31 @@
...
-        }
+        Self { status: Vec::new() }
     }
 
     /// Update the status
     pub fn update_status(&mut self, status: &str) {
-        self.status = status.to_string();
+        if self.status.len() >= 5 {
+            self.clear_status();
+        }
```
hannobraun, line 18: I think this implementation is a bit weird, as it results in lots of statuses being removed at once. Might be better to just remove the oldest status, until only 5 are left. Maybe [`VecDeque`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html) is a better type for this than `Vec`, as it provides convenient `push_back`/`pop_front` methods.

But that's something that can be fixed later.

hannobraun (approved):
Thanks for the latest changes, @devanlooches! And sorry for the delay; I was out and about for much of the weekend.

> Unfortunately I'm having trouble putting the window on the bottom right by default (and still having it able to be dragged). If anyone could help me with that it would be great.

Unfortunately I don't know either (I don't have much experience with egui). We can always improve it later! (And maybe the lower-right corner isn't even the best location.)

> One other thing I was thinking is it might be a good idea to keep the 5 status messages unless there is a compiler error in which case it might be a good idea to clear them in order to save space for the compiler error message (And also not show it more than once).

That sounds good!

I think it makes sense to distinguish between different kinds of status updates (info, error, etc.) and make decision based on what types we currently have.

---

I'm merging this now! There's still room for improvement, of course, but this is already much better than what we currently have. Thanks you again, @devanlooches!

crates/fj-interop/src/status_report.rs
devanlooches: No problem! Thanks for the feedback! Your right, VecDeque does look convenient. I'll look into it.

PR #934, opened 2022-08-09
Move the model traits into their own fj::models module

Michael-F-Bryan:
This is a follow-up to https://github.com/hannobraun/Fornjot/pull/885#issuecomment-1209235168 from @hannobraun. Originally I thought he meant to move the traits into a new crate (e.g. `fj-models`), but he was actually suggesting to move them out of the top-level namespace.

### DIFF
--- a/crates/fj-host/src/lib.rs
+++ b/crates/fj-host/src/lib.rs
@@ -389,7 +389,7 @@ pub enum Error {
     #[error("No model was registered")]
     NoModelRegistered,
 
-    /// An error was returned from [`fj::Model::shape()`].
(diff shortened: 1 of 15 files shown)


hannobraun (approved):
Thank you, @Michael-F-Bryan, much better!

One more nitpick: `abi` could be moved under `models`, and the `#[doc(hidden)]` could be removed. As I remarked, I have found libraries that use `#[doc(hidden)]` confusing to use in the past.

But that really is just a nitpick. It can be done whenever is convenient, and doesn't need to be done by you, unless you want to.

PR #935, opened 2022-08-09
Switch over to the abi-stable crate

Michael-F-Bryan:
This removes our hand-rolled FFI code in favour of using types and abstractions from the `abi-stable` crate.


hannobraun:
Hey @Michael-F-Bryan, thank you for doing this! This looks good, in principle, but that segmentation fault... I didn't notice anything suspicious. I wonder, if we could be doing something to cause that, or if it's a problem in `abi_stable`. I can't find any mentions of segmentation faults on their issue tracker.

hannobraun:
Wait, this triggers a memory. I think I already saw a segmentation fault some time ago, when I tried to use `abi_stable` for `PolyChain`, much like you did here. Back then, I think I couldn't rule out a mistake on our side, so I didn't open an issue in the `abi_stable` repository. I dropped that work shortly after, for lack of time.

Maybe it's worth setting up an isolated example, and running that with Miri. I tried to use Miri back then, but it was super-slow when running Fornjot as a whole, and tripped up on some unrelated errors in nalgebra, if I recall correctly.

hannobraun:
I got a stray segmentation fault last week, while [using Fornjot on another project](https://github.com/hannobraun/cnc-mill/tree/main/calculation). That was version 0.11.0, installed via `cargo install fj-app`.

I don't know how reproducible this is, and haven't had time to look into it. I plan to work some more on that project this week, and will follow up here once I have more information (and/or open a dedicated issue). But in any case, this points to `abi-stable` not being at fault.

hannobraun:
Update: I can't reproduce the segmentation fault that I saw earlier. I did notice that there had been a version mismatch between the Fornjot application and the `fj` crate, so maybe that was responsible.

hannobraun:
A new version of `abi-stable` was released since this attempt was made, with [lots of changes](https://github.com/rodrimati1992/abi_stable_crates/blob/master/Changelog.md#011). It might be worth trying this again!

hannobraun:
I've kept this pull request open, as I hoped that it could serve as a template for future work along those lines. However, due to the recent changes (see [A New Direction](https://www.fornjot.app/blog/a-new-direction/)), this is no longer in scope. Closing.

Thanks again for your work, @Michael-F-Bryan!

PR #945, opened 2022-08-12
Replace Vec with DecVeque and remove old statuses as new ones are added

devanlooches:


hannobraun (changes requested):
Thank you for this pull request, @devanlooches!

The change looks good, but the history is a bit confusing. This pull request contains many commits that have already been merged in earlier pull requests.

Could you to a `git rebase main`? In general, I prefer rebases over merges in most situations, as that results in a cleaner history that is easier to review.

devanlooches:
I agree sorry. Those commits are as a result of trying to sync the main branch with the fork by just pressing update :(

devanlooches:
oops I just did something stupid I think

devanlooches:
I think I deleted all of my changes. I'm going to remake this pull request cause I think I deleted all of my changes

hannobraun:
Oops, that sucks. Good thing, it wasn't too big of a change.

By the way, in cases like this, you can use `git reflog` to get your old commits back. In this case, it might be more work to figure out `git reflog` than it would be to redo the changes :smile:

devanlooches:
Cool thanks. I'll know for next time

PR #952, opened 2022-08-12
Replace Vec with VecDeque and remove status updates as they are added

devanlooches:


hannobraun (approved):
Thank you, @devanlooches, looks great!

PR #960, opened 2022-08-17
add compile time to status message

### LINKED ISSUE #936: Display model compile times
Fornjot models are defined in Rust, and Fornjot has support for compiling and loading those Rust-based models, as well as recompiling/reloading them, when they change. It used to be the case, that the full output from building the model was printed to stdout/stderr.

As of #911, this is no longer the case. There's a UI element now that displays updates on the model, as well as compiler errors, if there are any. Unfortunately, this has led to a loss of information. Specifically, the time it takes to compile a model is no longer displayed. This information is important though when developing Fornjot, as any regressions here affect the user experience negatively.

Every time a model is compiled, the time it took to compile it should be displayed together with the status update.

connor-lennox:
Addresses issue #936 .

I've changed how the command is being run from using `...status()` to `...output()`, which allows us to capture stdout/stderr earlier than in the error state. Quick parse over stderr in the success case gets us the time it took to compile, which I've inserted into the status update string to indicate how long it took to compile.

edit by @hannobraun, to close the issue:
Close #936

### DIFF
--- a/crates/fj-host/src/lib.rs
+++ b/crates/fj-host/src/lib.rs
@@ -25,6 +25,7 @@ use std::{
     ops::{Deref, DerefMut},
     path::{Path, PathBuf},
     process::Command,
+    str,
     sync::mpsc,
     thread,
 };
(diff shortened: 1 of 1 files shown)


hannobraun (approved):

Issue #986, opened 2022-08-24
Improve error message, if model can't be loaded


hannobraun:
When starting [`fj-app`](https://github.com/hannobraun/Fornjot/tree/main/crates/fj-app), you tell it which model to load by passing a command-line argument (`-m`/`--model`). If that model can't be found, this results in an error message. This error message can be hard to understand, and lacks information that could help to make sense of the problem.

Here's the error message that this currently results in:
```
Error: Failed to load model: models/cuboid

Caused by:
    0: I/O error while loading model
    1: No such file or directory (os error 2)
```

There should be a lot more information here, to help the user determine *why* the model couldn't be loaded:

- In addition to the relative path shown here, also show the absolute path, in case that makes the error more obvious.
- Explain, step-by-step, how the path came to be:
  1. The current working directory.
  2. The [default path from the configuration](https://github.com/hannobraun/Fornjot/blob/67cc32db7cf4b757413ebebf075c8330c90fa723/crates/fj-app/src/config.rs#L12).
  3. The model passed via `-m`/`--model`, or the [default model from the configuration](https://github.com/hannobraun/Fornjot/blob/67cc32db7cf4b757413ebebf075c8330c90fa723/crates/fj-app/src/config.rs#L13) as a fallback.

This information should help users recognize what went wrong.

The code that builds the path and creates the error message is located in `fj-app`:
https://github.com/hannobraun/Fornjot/blob/67cc32db7cf4b757413ebebf075c8330c90fa723/crates/fj-app/src/main.rs#L49-L59

I'm labeling this as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this is a fairly isolated change that doesn't require a lot of knowledge about the context.

Also see #987, which is related.

ArshErgon:
@hannobraun do you mean something like this?
```
Error: Failed to load model: models/cuboid
path: usr/Desktop/Fornjot/default_path/src/random.js
Caused by:
    0: I/O error while loading model 'random.js'
    1: Could not find "random.js" file inside 'src'
```

ArshErgon:
@hannobraun  do you mean something like this?

```
Error: Failed to load model: models/cuboid
path: usr/Desktop/default_path/model/file.js
Caused by:
    0: I/O error while loading model 'file.js'
    1: Could not find 'file.js' inside 'model', file not present

```

hannobraun:
Hey, @ArshErgon!

[...]

PR #1002, opened 2022-08-25
Error handling

ArshErgon:
updated the error message #986
this is how it will look to the console.
I just can't figure out a way to get the invalid model's name from the console, I think if I make a string of `path.display()` and then try to slice it I can get the command but it has to be a better way.
```
    Running `target/debug/fj-app -m testFile`
Error: 
Failed to load model: models/testFile inside default models directory: /home/arshergon/Desktop/openSource/Fornjot/models
                                            
Can mainly caused by: 
                                            
1.Model 'models/testFile' can not be found inside '/home/arshergon/Desktop/openSource/Fornjot/models' 
                                            
2.models/testFile can be mis-typed see inside /home/arshergon/Desktop/openSource/Fornjot/models for a match
                                            
3. Try to run 'cargo run -- -m cuboid' to know path is connected

Caused by:
    0: I/O error while loading model
    1: No such file or directory (os error 2)
```


hannobraun:
Thank you, @ArshErgon! I'll take a look later today.

crates/fj-app/src/main.rs
```diff
@@ -45,6 +45,9 @@ fn main() -> anyhow::Result<()> {
 
     let args = Args::parse();
     let config = Config::load()?;
+    
+    // finds the absolute path by the directory name
+    let abs_path_for_models = PathBuf::from("./models").canonicalize().unwrap();
```
hannobraun, line 50: The use of `./models` here is not correct. The `models/` directory has no special meaning for Fornjot. It just has a special meaning within this repository, because of the [configuration in `fj.toml`](https://github.com/hannobraun/Fornjot/blob/8fe463a9beba9c69c65a2ed61472a8c5db7c44c1/fj.toml#L3).

The correct path to call `canonicalize` on would be `path`, defined below, before `model` is pushed to it..

Also, this will panic, if a `models/` directory doesn't exist in the current working directory:

```
[hanno@grimbart models]$ cargo run -- -m cuboid
    Finished dev [unoptimized + debuginfo] target(s) in 0.22s
     Running `/home/hanno/Projects/main/fornjot/target/debug/fj-app -m cuboid`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', crates/fj-app/src/main.rs:50:72
```

This renders the point of creating a better error message moot :smile: 

To get a decent error message here, we'd need to replace the `.unwrap()` with a `.with_context(|| ...)?`, and create an error message within the closure, like below.

crates/fj-app/src/main.rs
```diff
@@ -55,13 +58,25 @@ fn main() -> anyhow::Result<()> {
...
+                                            \nCan mainly caused by: 
+                                            \n1. Model '{}' can not be found inside '{}' 
+                                            \n2. {} can be mis-typed see inside {} for a match
+                                            \n3. Try to run 'cargo run -- -m cuboid' to know path is connected", 
+                                            abs_path_for_models.display(), 
+                                            path.display(), 
+                                            abs_path_for_models.display(),
+                                            path.display(),
+                                            abs_path_for_models.display(),
+                                        );
```
hannobraun, line 71: There's a trick to make this kind of thing a bit easier: you can pass each argument only once, and refer to the arguments from the format string by position, so you can use it multiple times. Something like this:

``` rust
format!("{0}, {1}, {0}, {1}, arg0, arg1);
```

This is documented here: https://doc.rust-lang.org/stable/std/fmt/index.html#positional-parameters

Also item 3. in the error message is too specific. It only makes sense within this repository, and won't make sense to anyone who has installed `fj-app` via `cargo install` or by downloading a release binary.

hannobraun (changes requested):
Thank you for this pull request, @ArshErgon! I found some problems and left some comments.

Also, the CI build is failing due to formatting. Running `cargo fmt` should fix this (see [formatting section in `CONTRIBUTING.md`](https://github.com/hannobraun/Fornjot/blob/8fe463a9beba9c69c65a2ed61472a8c5db7c44c1/CONTRIBUTING.md#formatting) for more information).

crates/fj-app/src/main.rs
ArshErgon: @hannobraun yeah, I updated the format and the `abs_path` but I can't figure out how to get the invalid file name? I know its inside the `model` variable everytime if I try to get it, its return back with ` "models/InvalidFile" `

hannobraun: Thanks for following up, @ArshErgon! I'll check back and try to answer your question as soon as I can, but I'm on my way out of the house right now. I might not get a chance too look into this until Monday.

ArshErgon: @hannobraun I figure out, I reformatted the code and added the code.

hannobraun (changes requested):
Thank you for the updates, @ArshErgon! And sorry for the delay. I'm back in the office now, so I finally had a chance to take a look at this again.

I tested this change and found some problems. I've added comments with suggestions.

crates/fj-app/src/main.rs
```diff
@@ -43,22 +44,32 @@ fn main() -> anyhow::Result<()> {
         .event_format(format().pretty())
         .init();
 
+    let file_name: Vec<String> = env::args().collect();
```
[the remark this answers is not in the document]
ArshErgon: well if I the model which we have to run will always be in the last, the getting the model from the last would work

ArshErgon:
@hannobraun I re-edit the code, but finding model I can't figure it out, Rust take an average time of 4 to 7 mins to compile and sometimes it freezes my system, I can't run many tests on it.

hannobraun:
@ArshErgon 
> I re-edit the code, but finding model I can't figure it out,

As I said earlier, I think you just need to use the `model` variable:
https://github.com/hannobraun/Fornjot/blob/5d599626164b62a49ec95da0053633b2f0e42f0a/crates/fj-app/src/main.rs#L56-L61


> Rust take an average time of 4 to 7 mins to compile and sometimes it freezes my system, I can't run many tests on it.

Oh, wow, I didn't realize it could get this bad. Hard to get work done under these circumstances!

ArshErgon:
@hannobraun yeah, I have tried the attributes of model but I couldn't find one which get the thing done.

hannobraun:
Hmm, I'm not sure what to do then. Maybe I can look into it myself at some point, but I'm pretty busy right now.

hannobraun:
Since this needs more work before it's ready to merge, I'm marking it as a draft, for the time being.

hannobraun:
Closing in favor of #1154.

Issue #1020, opened 2022-08-30
Bound all edges by vertices


hannobraun:
`Edge`s are defined by two things:

- The (infinite) curve they are on.
- The vertices that bound the (finite) edge on that curve.

Except, this is not completely true. There are also continuous edges, which are edges that connect to themselves and don't have any vertices to bound them. They can only exist on continuous curves (curves that connect to themselves; right now only circles are supported).

This exception causes quite a few problems:

- There is code all over `fj-kernel` that handles this special case in all kinds of circumstances.
- Sweeping of continuous curves is the last piece of code that prevents us from removing triangle representation (#97).
- Approximation doesn't support continuous faces (#250).

I believe that all of these problems can be solved by abolishing the concept of continuous curves (and, by extension, continuous surfaces and faces) and making sure that all `Edge`s are bound by vertices (and, by extension, all faces are bound by edges). This promises a lot of simplification, which would pay dividends (in saved work) pretty quickly.

This isn't a new idea. In fact, I believe this is a pretty normal way to do CAD kernels, and I mentioned this previously in #250. I didn't want to do it earlier, because a) I'm a bit of an idiot, and b) I didn't see a clear way to do this. The kernel data structures have improved a lot over the last months though (#691 was a big one, in this context), so it should now be possible to implement this *relatively* easily.

Over the last week or so, I've been working on this, and I've made big strides towards making it possible, by updating the approximation code to support this new notion of edges (#1003, #1006, #1011, #1012, #1013). I also have a work-in-progress branch that starts making the actual change to `Edge` (#1019), but unfortunately this causes issues in the sweep algorithm.

For now, I will turn my attention to other things, but I expect that I'll have the opportunity to address this soon. The sweep algorithm is also blocking #993, which in turn is blocking my top priority, boolean operations (#42, #43, #44). Once the sweep algorithm is sorted out, this will hopefully become easy to finish.

hannobraun:
I forgot to update this issue: I've been working on this for about a week. See https://github.com/hannobraun/Fornjot/issues/993#issuecomment-1234425471 for context.

hannobraun:
My WIP implementation in #1019 is pretty much finished, but there's a bug in the approximation code that causes invalid triangle meshes to be created, for some models. The root cause for that is a reversed curve that generates a slightly different approximation than its non-reversed but coincident counterpart.

[...]

Issue #993, opened 2022-08-24
There is no way to determine whether two edges are equal


hannobraun:
When dealing with edges in the kernel (instances of struct [`Edge`](https://github.com/hannobraun/Fornjot/blob/7751b8fe1a38cc63457079c1ed0187779c13f4a2/crates/fj-kernel/src/objects/edge.rs#L7-L12), to be precise), there is no way to tell whether two edges are actually the same. This hasn't been a problem so far, but it is getting in the way of finishing some intersection algorithms I'm currently writing for #42.

This problem consists of multiple sub-problems:

1. `Edge` is defined in terms of local surface coordinates, and there is no `GlobalEdge` that could be used to determine whether two `Edge`s are the same in global space. This is actually quite trivial to address (I have the solution in a local branch), but by itself, this doesn't help.
2. `Edge`s are actually half-edges, as in, they are directed. So even if we had a `GlobalEdge`, we'd still be dealing with the problem that two `GlobalEdge`s could be coincident, but still defined differently. This is related to #695.
3. The sweep algorithm produces `Edge` instances that are, in principle, the same, but they are defined differently. Both in terms of their local coordinates (which can't be avoided), but also in the direction of their global ones. The shared edge of two neighboring side faces isn't actually shared; it's basically two different edges that are coincident, but point in different directions.

The last problem is the most difficult, probably. The first two just require a bit of re-shuffling within the kernel code, but the creation of side faces in the sweep algorithm has been a problem again and again, since that algorithm exists. Fortunately time has gone on, and I have new ideas on how to finally do this right. I'll report back once I had the chance to do a bit of experimentation.

hannobraun:
I'm currently working on this, as it's blocking #42 (which is my top priority).

hannobraun:
I've been working on this for a bit, so here's an update: Sub-problem 1. from my original comment here has been addressed (#999). As I said, this doesn't do a whole lot by itself, and I'm still working on the other sub-problems.

[...]

PR #1031, opened 2022-09-02
Expand validation code

hannobraun:
I'm currently working on cleaning up the sweep algorithm, so I can make some progress on #993 and #1020. This work has unearthed many ways to mess up the data structures. This new validation code catches the issues I found, and should make things much easier in the future.


hannobraun:
Oh, wow, looks like the new curve coherence validation uncovered a bug in the existing sweep code! I wrote it to get the issues of my rewritten sweep code under control, but uncovering this bug is a nice bonus.

I guess I'll have to merge the rewritten sweep code first, before I can merge this PR.

Issue #987, opened 2022-08-24
Only use default path, if `--model` doesn't actually point to a model


hannobraun:
When starting [`fj-app`](https://github.com/hannobraun/Fornjot/tree/main/crates/fj-app), you tell it which model to load by passing a command-line argument (`-m`/`--model`). The argument passed isn't always interpreted as a relative path though: There is [configuration](https://github.com/hannobraun/Fornjot/blob/67cc32db7cf4b757413ebebf075c8330c90fa723/crates/fj-app/src/config.rs#L11-L14) that can affect from where the model is loaded, namely the `default_path` key. See the [`fj.toml`](https://github.com/hannobraun/Fornjot/blob/main/fj.toml) from this repository for an example.

This can be confusing, as the user might not be aware that a configuration file exists, and that it affects the loading behavior. I think the following behavior would be clearer:

1. If the argument passed via `-m`/`--model` is a path, either absolute, or relative to the current working directory, where a Fornjot model is found, load that model.
2. Only if no model is found, fall back to using `default_path`.

The code responsible for the current behavior lives in `fj-app`:
https://github.com/hannobraun/Fornjot/blob/67cc32db7cf4b757413ebebf075c8330c90fa723/crates/fj-app/src/main.rs#L49-L59

Some thoughts on the implementation:

- I can think of two ways to determine whether a directory contains a model:
  1. Do a check for `Cargo.toml` (maybe in [`fj_host::Model::from_path`](https://github.com/hannobraun/Fornjot/blob/67cc32db7cf4b757413ebebf075c8330c90fa723/crates/fj-host/src/lib.rs#L53-L76)) and return an error if it doesn't exist.
  2. Try to load the model at that path normally, and retry with the fallback path, if there's an error that indicates the model doesn't exist. This method could be more reliable, but would likely require a rework of [`fj_host::Error`](https://github.com/hannobraun/Fornjot/blob/67cc32db7cf4b757413ebebf075c8330c90fa723/crates/fj-host/src/lib.rs#L380-L433) to make it possible/easy to determine that a model doesn't exist there, as opposed to an error occurring with an existing model.
[...]

PR #1069, opened 2022-09-11
Refactor model import.

whereistejas:
Fixes: #987.

Also see, https://github.com/hannobraun/Fornjot/issues/987#issuecomment-1242987267

### DIFF
--- a/crates/fj-app/src/main.rs
+++ b/crates/fj-app/src/main.rs
@@ -55,25 +55,29 @@ fn main() -> anyhow::Result<()> {
     })?;
     path.push(model);
 
-    let model = Model::from_path(path.clone())
-        .with_context(|| format!("Failed to load model: {}", path.display()))?;
     let parameters = args.parameters.unwrap_or_else(Parameters::empty);
-
     let shape_processor = ShapeProcessor {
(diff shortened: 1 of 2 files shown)


hannobraun:
Thank you for the pull request, @whereistejas! I've taken a quick look, and think this looks reasonable. I'll do a proper review later today, or possibly tomorrow.

Please note that there are two CI failures. A Clippy warning that looks trivial to fix, and a model loading error.

hannobraun:
Hey @whereistejas, I've taken another quick look. I still think the new API looks good, but there seems to be something fishy going on with the path. `cargo run` from the repository root fails with an error, as does trying to load any specific model (e.g. `cargo run -- -m test`).

Please let me know when this is ready for review, or if you need any help!

hannobraun:
Hey @whereistejas, are you still interested in working on this?

whereistejas:
hey @hannobraun sorry for late reply! yes, I totally intend to see this to the end. expect some new commits very soon! after having gone through the code more thoroughly, I don't think changing the structure of the code adds a lot of value, maybe we can just rename the methods.

Issue #917, opened 2022-08-05
Review UI, decide what needs to be done


hannobraun:
The UI is in a rather experimental state. There are quite a few issues that track possible improvements (see https://github.com/hannobraun/Fornjot/labels/topic%3A%20ui), but so far no comprehensive review has been done.

It would be good to take a closer look, figure out what's missing or doesn't work well, compare that against the already open issues, and open new issues as required.

hannobraun:
This issue is no longer applicable. The affected code [has been removed](https://github.com/hannobraun/fornjot/pull/1822). See [A New Direction](https://www.fornjot.app/blog/a-new-direction/) for context.

PR #1095, opened 2022-09-15
Fix #1015 start gui independent of model

### LINKED ISSUE #1015: GUI should start immediately, without waiting for model to be loaded
Right now, when starting the Fornjot app, first the model is being loaded, then the GUI is shown with the loaded model. This causes problems:

- Sometimes loading the model takes longer, for some reason. Then it takes a while for the GUI to show.
- If there's an error loading the model (for example a compile error in the model code), the GUI doesn't load at all.

This is very confusing. I think the solution is to load the GUI first (not showing any model), then show the model once it is ready. I don't think this will be hard to achieve, but currently code in `fj-window`/`fj-viewer` assumes that there is always a model, so this would have to be adapted.

I'm classifying this as https://github.com/hannobraun/Fornjot/labels/type%3A%20bug, since this is such a confusing problem. Especially if the model doesn't load at all, since there is no GUI to show the error then.

payload:
Fixes #1015
Rendering the gui depends on a camera being there but the camera was only initialized when the model watcher loads something from the model.
Now there is always a camera and in case there is a model watcher the camera is reset on load once.

In main loading a model was always assumed. Since a model is now optional upon start of the app, we have following cases.
If there is a model parameter or config, create a Model::from this model parameter. If this fails already, exit.
If there is an export parameter, check that there is a model. If not exit with "no model specified". Else load once, process and export. Done.
If there is a model, create a watcher and load. Run the GUI.

If there is no model, run the GUI without watcher. This is kind of useless now until we can select and load a model from the GUI.

### DIFF
--- a/crates/fj-app/src/main.rs
+++ b/crates/fj-app/src/main.rs
@@ -73,6 +73,7 @@ fn main() -> anyhow::Result<()> {
     }
 
     let watcher = model.load_and_watch(parameters)?;
+    // run(Some(watcher), shape_processor, status)?;
     run(None, shape_processor, status)?;
 
     Ok(())
--- a/crates/fj-app/src/main.rs
+++ b/crates/fj-app/src/main.rs
@@ -73,7 +73,7 @@ fn main() -> anyhow::Result<()> {
     }
 
     let watcher = model.load_and_watch(parameters)?;
-    run(watcher, shape_processor, status)?;
+    run(None, shape_processor, status)?;
 
     Ok(())
 }
--- a/crates/fj-app/src/main.rs
+++ b/crates/fj-app/src/main.rs
@@ -46,33 +46,38 @@ fn main() -> anyhow::Result<()> {
     let args = Args::parse();
     let config = Config::load()?;
 
-    let mut path = config.default_path.unwrap_or_else(|| PathBuf::from(""));
-    let model = args.model.or(config.default_model).ok_or_else(|| {
-        anyhow!(
-            "No model specified, and no default model configured.\n\
-                Specify a model by passing `--model path/to/model`."
-        )
-    })?;
-    path.push(model);
-
-    let model = Model::from_path(path.clone())
-        .with_context(|| format!("Failed to load model: {}", path.display()))?;
+    let path = config.default_path.unwrap_or_else(|| PathBuf::from(""));
     let parameters = args.parameters.unwrap_or_else(Parameters::empty);
-
     let shape_processor = ShapeProcessor {
         tolerance: args.tolerance,
     };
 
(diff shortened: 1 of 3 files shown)


payload:
This PR is a first attempt. I am happy to polish it a little bit upon some discussion.

crates/fj-window/src/run.rs
```diff
@@ -46,39 +46,43 @@ pub fn run(
...
+                            new_shape.aabb,
+                        );
 
-                    if camera.is_none() {
-                        camera = Some(Camera::new(&new_shape.aabb));
-                    }
+                        if camera_update_once {
+                            camera_update_once = false;
+                            camera = Camera::new(&new_shape.aabb);
+                        }
```
hannobraun, line 68: I'm not sure I like `camera_update_once`. It seems like the kind of thing that could possibly turn into a bug, as things get shuffled around over time. But it *is* nicer, than having `if let Some(camera)` everywhere.

I think maybe `Camera` is no longer well-suited to this new situation. We could make the constructor parameter-less, just creating reasonable defaults there, and have some kind of internal flag that makes sure that `update_planes` does what `Camera::new` currently does when first called.

Then we could just create the camera using `Camera::new` above and do an unconditional `update_planes` here, instead of down there right before the rendering. But that would be work for a follow-up pull request (and if you're inclined to do it, I'd be happy to merge it).

crates/fj-app/src/main.rs
```diff
@@ -46,34 +46,46 @@ fn main() -> anyhow::Result<()> {
...
     }
 
-    let watcher = model.load_and_watch(parameters)?;
-    run(watcher, shape_processor, status)?;
+    if let Some(model) = model {
+        let watcher = model.load_and_watch(parameters)?;
+        run(Some(watcher), shape_processor, status)?;
+    } else {
+        run(None, shape_processor, status)?;
+    }
```
hannobraun, line 88: This could be made a bit more compact (but possible less clear) using [`Option::map`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map). Not necessary to change anything, and I don't know if it would even be better. Just noting the possibility, because it stuck out to me.

hannobraun (approved):
Thank you, @payload, this looks great!

I've left some comments with suggestions, but this is all minor stuff, as you didn't give me anything to *actually* complain about.

> If there is no model, run the GUI without watcher. This is kind of useless now until we can select and load a model from the GUI.

This could be confusing to users, but that's acceptable for now. Just one more thing to take care of in #917.

---

Merging this now. Feel free to address any of the comments in follow-up PRs, or not, as you prefer.

Issue #1079, opened 2022-09-13
Consider revamping `GlobalCurve`


hannobraun:
`Curve` is defined in surface (2D) space while `GlobalCurve` is defined in global (3D) space. Hence, `Curve` is used to convert curve (1D) coordinates to surface coordinates, while `GlobalCurve` is used to convert those curve coordinates to global coordinates.

This is working well enough for now, but I don't think it will be a suitable solution going forward.

## Problem 1: Redundancy between `Curve`/`GlobalCurve`

`GlobalCurve` isn't actually needed for for generating global coordinates. `Curve` references the `Surface` it is defined on, and that `Surface` can be used to convert the surface coordinates that `Curve` generates into global coordinates, bypassing `GlobalCurve`.

So, as far as computing approximations goes, we don't need `GlobalCurve`. This isn't really much of a problem, but it's supporting evidence for what is going to be my proposal.

## Interlude 1: Let's cache curve approximations

If you've been following carefully and have sufficient knowledge of the Fornjot kernel, you might note that what I've just said wouldn't actually work. It could result in slightly different 3D points being generated for `Curve` approximations that are supposed to be the same, thus resulting in an invalid mesh being generated. (If you actually noted that, please send more pull requests. I can use the help. :smile:)

Right now, `Curve` approximation goes through `GlobalCurve`, and we're being really careful about approximations of should-be-identical `GlobalCurve`s producing the same points (even if those `GlobalCurve`s point in different directions). If we started going through both the `Curve` and the `Surface` to create a 3D approximation, then no amount of being really careful would prevent invalid meshes.

This "being really careful" business is probably already a warning sign that there's a deeper problem not being addressed. So let's just remove the need for being careful in the first place. When approximating a `Curve`, let's cache the approximation, using the `GlobalCurve` as a key.

(Actually, we can only cache the curve-space and global-space points generated for the approximation. We'll need to re-compute the surface points per `Curve`, but that's just an implementation detail and not a problem.)

This might have the side effect of saving redundant computations, leading to better performance. But right now, these re-computations don't seem to be a performance problem, so whatever.

## Problem 2: `CurveKind` won't cut it going forward

Both `Curve` and `GlobalCurve` own a `CurveKind`, which defines the actual geometry of the curve. To accommodate both `Curve` and `GlobalCurve`, `CurveKind` is generic over its dimension. `Curve` owns a `CurveKind<2>`, `GlobalCurve` owns a `CurveKind<3>`.

[...]

Issue #1021, opened 2022-08-31
Implement lightweight, centralized object storage


hannobraun:
Not too long ago, we used to have the `Shape` data structure which represented a single shape. It stored the shape's objects and provided a whole lot of infrastructure to manage them. `Shape` was overly complex, and was removed after simpler ways to do the things it did were found (#697). Since then, we no longer have any form of centralized storage of objects.

Even back then, I suspected that we'd need to grow back some form of centralized object storage eventually. Although much lighter-weight, and without the problematic parts that `Shape` had. I've been thinking about this in the back of my mind, and this issue is the result of that.

## Why do we need centralized object storage?

So why do we need it? Centralized object storage gives us one thing we can't get any other way: We'll be able to refer to existing objects using a handle, which is kind of a reference to the object in the store.

But how is that an advantage? There are two ways that I can see:

1. It's lighter-weight. In principle, objects reference other objects. In practice, right now, it's not a reference. Instead, the referencing object owns a copy of the referenced object, resulting in much duplication.
2. It allows to distinguish between object identity and object equality. This allows for more stringent data structures that could prevent some bugs.

Advantage 1. doesn't *really* matter right now. Yes, it would be nice to slim down the data structures somewhat (I don't have numbers, but I suspect that they are many times bigger than they'd need to be), but that doesn't cause any problems right now. Long-term, it might make a difference in performance.

Advantage 2. is a bit more substantial, in that it gives us back a capability that we lost when removing `Shape`. Back then, I said that object identity wasn't as important as I thought it would be, but I'm no longer sure. Accidentally creating an object that is identical to an already existing one, when you should just use the already existing one instead, might be a sign of sloppy code that could cause other problems.

Although, to be honest, I can't point to evidence of that right now. This is just an impression that has grown over time, as I'm working on the kernel code. But in any case, centralized object storage gives us the *option* of being strict about object identity, while still allowing for escape valves where necessary (in the form of a method like `store.get_handle_for_or_insert(object)`).

## How is that going to be different from `Shape`?

How can we make sure not to repeat the same mistakes though? Here's where I see the differences between the new object store and `Shape`:

- It's going to be a global store for all shapes, instead of a per-shape store. This avoids most of the redundancy that was inherent to `Shape`.
- Most of `Shape`'s problematic use cases have already been solved in other ways. It simply won't be necessary to rebuild them in the new API.
- The new API will strictly be an append-only data structure.
[...]

PR #1135, opened 2022-09-23
Convert edge build API into partial edge API

hannobraun:
This is another step towards addressing the object identity problems that are currently holding up #1079.


hannobraun:
This is failing the CI build. It's a validation error, because the half-edges of a cycle do not connect (caught by the export validator in the star model). There's only a minuscule difference between vertices that are supposed to be the same, so it's a floating-point accuracy issue. It seems to start with this commit: https://github.com/hannobraun/Fornjot/pull/1135/commits/0c7c508a45f131da4eb559f36980525baecd4241

I don't think that this commit actually introduces the bug. I might be mistaken, but I think it just exposes a bug that was already there. The vertices were already computed redundantly, and it just happened to work. The new code just happens to compute slightly different vertices redundantly, therefore exposing the bug.

This is exactly the kind of bug that would no longer be possible, if validation were based on object identity, not object equality. Then it would *always* cause a validation error, hence making it impossible for such a bug to hide. It's one of the reasons I want #1021, and why I'm doing all this work with the builder / partial object API.

I'll try to locate and fix the buggy piece of code, but I might not be able to finish that today. If so, I'll leave this open until next week.

hannobraun:
I've located the bug and pushed a commit that fixes it before it is uncovered (meaning all the commits in this pull request should pass the full test suite). Here it is: https://github.com/hannobraun/Fornjot/pull/1135/commits/73bdcdb262078f226c512e91c670e16441b53202

The fixed code isn't exactly beautiful, but it's fine for now. It's in a builder I have yet to update, so I'll have an opportunity to clean it up soon.

PR #1154, opened 2022-09-29
Error handling

### LINKED ISSUE #986: Improve error message, if model can't be loaded

ArshErgon:
for the issue #986 I have re-edited the error the Fornjot application.
Do review the PR

edit by @hannobraun:
Close #986 (to close the issue when this is merged)

### DIFF
--- a/crates/fj-app/src/main.rs
+++ b/crates/fj-app/src/main.rs
@@ -51,7 +51,7 @@ fn main() -> anyhow::Result<()> {
         tolerance: args.tolerance,
     };
 
-    let path_of_model = path.canonicalize().unwrap();
+    let path_of_model = path.canonicalize().unwrap_or_default();
 
     let model = if let Some(model) =
         args.model.or(config.default_model).as_ref()
@@ -59,9 +59,15 @@ fn main() -> anyhow::Result<()> {
         let mut model_path = path;
         model_path.push(model);
         Some(Model::from_path(model_path.clone()).with_context(|| {
-            format!(
+            if path_of_model.as_os_str().is_empty() {
+                format!(
+                    "Model is not defined, can't find model defined inside the default-model also, add model like \n cargo run -- -m {}", model.display()
+                )
+            } else {
+                format!(
                 "Failed to load model: {0}\ninside default models directory: '{1}'\nCan mainly caused by: \n1. Model '{2}' can not be found inside '{1}'\n2.'{2}' can be mis-typed see inside '{1}' for a match\n3. Define model is '{2}' couldn\'t be found ((defined in command-line arguments))", model_path.display(), path_of_model.display(), model.display()
             )
+        }
         })?)
     } else {
         None
--- a/crates/fj-app/src/main.rs
+++ b/crates/fj-app/src/main.rs
@@ -58,7 +58,7 @@ fn main() -> anyhow::Result<()> {
     };
 
     let path_of_model = path.canonicalize().unwrap();
-    let new_error_message = format!("inside default models directory: '{0}'\nCan mainly caused by: \n1. Model '{1}' can not be found inside '{0}'\n2.'{1}' can be mis-typed see inside '{0}' for a match \n3. Define model is'{1}' couldn\'t be found ((defined in command-line arguments))", path_of_model.display(), fetch_model_name);
+    let new_error_message = format!("inside default models directory: '{0}'\nCan mainly caused by: \n1. Model '{1}' can not be found inside '{0}'\n2.'{1}' can be mis-typed see inside '{0}' for a match \n3. Define model is '{1}' couldn\'t be found ((defined in command-line arguments))", path_of_model.display(), fetch_model_name);
     let model = if let Some(model) = args.model.or(config.default_model) {
         let mut model_path = path;
         model_path.push(model);
--- a/crates/fj-app/src/main.rs
+++ b/crates/fj-app/src/main.rs
@@ -45,18 +45,22 @@ fn main() -> anyhow::Result<()> {
 
     let args = Args::parse();
     let config = Config::load()?;
-
+    // model name defined here.
+    let model_name_from_cli = args.model.as_ref().expect("REASON").display();
+    // path for the models
     let path = config.default_path.unwrap_or_else(|| PathBuf::from(""));
     let parameters = args.parameters.unwrap_or_else(Parameters::empty);
     let shape_processor = ShapeProcessor {
         tolerance: args.tolerance,
     };
 
+    let path_of_model =  path.canonicalize().unwrap();
+    let new_error_message = format!("inside default models directory: '{0}'\nCan mainly caused by: \n1. Model '{1}' can not be found inside '{0}'\n2.'{1}' can be mis-typed see inside '{0}' for a match \n3.'{1}' couldn\'t be found", path_of_model.display(), model_name_from_cli);
     let model = if let Some(model) = args.model.or(config.default_model) {
         let mut model_path = path;
         model_path.push(model);
         Some(Model::from_path(model_path.clone()).with_context(|| {
-            format!("Failed to load model: {}", model_path.display())
+            format!("Failed to load model: {} {new_error_message}", model_path.display())
         })?)
     } else {
         None
--- a/crates/fj-app/src/main.rs
+++ b/crates/fj-app/src/main.rs
@@ -45,12 +45,6 @@ fn main() -> anyhow::Result<()> {
 
     let args = Args::parse();
     let config = Config::load()?;
-    // model name defined here.
-    let fetch_model_name =
-        (args.model.as_ref().or(config.default_model.as_ref()))
-            .unwrap()
-            .display();
-    // path for the models
     let path = config.default_path.unwrap_or_else(|| PathBuf::from(""));
     let parameters = args.parameters.unwrap_or_else(Parameters::empty);
     let shape_processor = ShapeProcessor {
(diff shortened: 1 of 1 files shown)


hannobraun (changes requested):
Thank you for the new pull request @ArshErgon! Unfortunately it still doesn't work. Here's what I get when running the Fornjot app from the repository root, without arguments:

```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/fj-app`
thread 'main' panicked at 'REASON', crates/fj-app/src/main.rs:49:51
```

We can't rely on the model being passed on the command line, as a default model can be specified in the configuration. This is why the code here looks at both the arguments and the configuration:
https://github.com/hannobraun/Fornjot/blob/8ece6e23c90b51495c192dd393372ae1b1910a8b/crates/fj-app/src/main.rs#L55

I think it should be possible to get the path for the error message from the `model_path` variable:
https://github.com/hannobraun/Fornjot/blob/8ece6e23c90b51495c192dd393372ae1b1910a8b/crates/fj-app/src/main.rs#L56-L57

ArshErgon:
Hey @hannobraun well if I use `model_path` it will return as `model/testFile`, so this will it work for error message?

hannobraun:
> if I use `model_path` it will return as `model/testFile`, so this will it work for error message?

Oh, wait, I think I said the wrong thing? The model name should be in the `model` variable, not `model_path`.

ArshErgon:
okay, but can you tell me is anything in rust which can tell me attributes of a particular variable? like in python we have `print(dir(object))` or maybe you could recommend me some good VS code extension, at now I only have `rust-analyzer` its pretty hard for me to know what methods does a particular variable using

hannobraun:
> okay, but can you tell me is anything in rust which can tell me attributes of a particular variable? like in python we have `print(dir(object))`

There's nothing in Rust that would let you print the methods that a type has, as far as I know. This makes sense, as Rust is a statically compiled language (unlike Python), so this kind of data is no longer available at runtime.

You can print actual data though, using [`dgb!`](https://doc.rust-lang.org/std/macro.dbg.html). There are other ways too, for example [`print!`](https://doc.rust-lang.org/std/macro.print.html)/[`println!`](https://doc.rust-lang.org/std/macro.println.html).


> or maybe you could recommend me some good VS code extension, at now I only have `rust-analyzer`

rust-analyzer is definitely the best Rust extension for VS code, but it's not well-suited to figure out which methods a type has. The best way to figure that out is to look at the documentation.


> its pretty hard for me to know what methods does a particular variable using

For types from the standard library, I recommend the [official documentation](https://doc.rust-lang.org/std/index.html). If you're using a type from a 3rd-part library, you can look up that library on [crates.io](https://crates.io/). Most have a link to documentation on the right. Most documentation is hosted on [docs.rs](https://docs.rs/), so you can also search there directly.

If you want to see the source code of a type, you can jump to it using rust-analyzer, using *Right Click* -> *Go to Definition*, or by pressing *F12* while the cursor is on top of the type name.

ArshErgon:
@hannobraun is this you want? `cargo run` isn't panicking anymore

crates/fj-app/src/main.rs
```diff
@@ -45,18 +45,28 @@ fn main() -> anyhow::Result<()> {
 
     let args = Args::parse();
     let config = Config::load()?;
-
+    // model name defined here.
+    let fetch_model_name =
+        (args.model.as_ref().or(config.default_model.as_ref()))
+            .unwrap()
```
hannobraun, line 51: This breaks when no model is defined at all, so neither via a command-line argument, nor as a default model in the configuration. This is a case that is otherwise handled, see the [`if let` in line 62](https://github.com/hannobraun/Fornjot/blob/45abaf5feff821e311406789043133598221a1ed/crates/fj-app/src/main.rs#L62) and [its `else` in lines 96-98](https://github.com/hannobraun/Fornjot/blob/45abaf5feff821e311406789043133598221a1ed/crates/fj-app/src/main.rs#L96-L98).

This can be reproduced like this:
1. While in your branch, install `fj-app` by running this from the repository root: `cargo install --path crates/fj-app/`
2. Run this from any directory that has no Fornjot model, nor Fornjot configurat: `fj-app`

This results in the following error:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', crates/fj-app/src/main.rs:51:14
```

Your code here duplicates the logic from line 62, which is a warning sign. Whenever you duplicate code like that, there's a danger that the two versions are different and one doesn't work correctly (as is the case here). Even if both work correctly, sooner or later one of the versions will be changed, and whoever does that might forget to update the other one.

The solution is to re-use the code that is already there. Instead of creating a new variable (`fetch_model_name`) here, why not move the code that formats the new error message into the `if let` (line 63 and following)? Then it can use the `model` variable that is already there.

hannobraun (changes requested):
Thank you for keeping at it, @ArshErgon!

Unfortunately it now breaks in a different way. I've left a comment to explain that.

ArshErgon:
@hannobraun I actually laughed when I read `it now breaks in different way` will see what I can do. thanks for guiding me so far.

crates/fj-app/src/main.rs
```diff
@@ -45,18 +45,29 @@ fn main() -> anyhow::Result<()> {
...
+    let path_of_model = path.canonicalize().unwrap_or_default();
+
+    let model = if let Some(model) =
+        args.model.or(config.default_model).as_ref()
+    {
         let mut model_path = path;
         model_path.push(model);
         Some(Model::from_path(model_path.clone()).with_context(|| {
-            format!("Failed to load model: {}", model_path.display())
+            if path_of_model.as_os_str().is_empty() {
```
hannobraun, line 62: Under which circumstances can `path_of_model` be empty? I suspect this can only happen, if `path.canonicalize()` above returns an `Err`.

If that is true, we can make things a bit more direct, doing something like this:

```
let abs_path = if let Ok(path_of_model) = path.canonicalize() {
    // insert error message using `path_of_model` here
} else {
    // insert error message that doesn't use `path_of_model` here
}
```

As a side note, this would be a prime use case [`let-else`](https://rust-lang.github.io/rfcs/3137-let-else.html). This feature isn't stable yet, but should become available soon.

crates/fj-app/src/main.rs
```diff
@@ -45,18 +45,29 @@ fn main() -> anyhow::Result<()> {
...
+                format!(
+                    "Model is not defined, can't find model defined inside the default-model also, add model like \n cargo run -- -m {}", model.display()
+                )
```
hannobraun, line 65: That error message could use some more detail. It doesn't say where it searched for the model, unlike the more detailed error message below.

crates/fj-app/src/main.rs
```diff
@@ -45,18 +45,29 @@ fn main() -> anyhow::Result<()> {
...
+            } else {
+                format!(
+                "Failed to load model: {0}\ninside default models directory: '{1}'\nCan mainly caused by: \n1. Model '{2}' can not be found inside '{1}'\n2.'{2}' can be mis-typed see inside '{1}' for a match\n3. Define model is '{2}' couldn\'t be found ((defined in command-line arguments))", model_path.display(), path_of_model.display(), model.display()
+            )
```
hannobraun, line 69: Having the error message in one long line like this can be a bit hard to read, so here's a tip on formatting strings: You can add a `\` to the end of a line, which will make the parser ignore everything until the next non-whitespace character. Using that trick, you can make it look like this:

```suggestion
                format!(
                    "Failed to load model: {0}\n\
                    inside default models directory: '{1}'\n\
                    Can mainly caused by: \n\
                    1. Model '{2}' can not be found inside '{1}'\n\
                    2.'{2}' can be mis-typed see inside '{1}' for a match\n\
                    3. Define model is '{2}' couldn\'t be found ((defined in command-line arguments))",
                    model_path.display(),
                    path_of_model.display(),
                    model.display()
            )
```

hannobraun (approved):
Thank you for the latest changes, @ArshErgon! Everything works great now, as far as I can tell. Thank you for keeping at it!

I've left some comments with suggestions, but this is fine as-is, and I'll be merging it now. Feel free to address these comments in a follow-up pull request, if you want to (or not, if you don't).

crates/fj-app/src/main.rs
ArshErgon: @hannobraun I will truly do another PR again,
I also tried to use that `if let` case but I was running it with `Some` and couldn't figure it out, and settle with using the simple `if else`.

crates/fj-app/src/main.rs
ArshErgon: sure will update it with something bit more reasonable.

crates/fj-app/src/main.rs
hannobraun: > I will truly do another PR again,

Thank you, that's great!

---

In case you're wondering how to figure out these kinds of things, here's where to find the necessary information in the standard library API reference:

If you check out the documentation, you can see that [`Path::canonicalize`](https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize) returns a `Result` (specifically an [`io::Result`](https://doc.rust-lang.org/std/io/type.Result.html), which is defined in terms of [`result::Result`](https://doc.rust-lang.org/std/result/enum.Result.html)). `Result` has `Ok` and `Err` variants, which you can use with `if let` or `match`.

`Some`, on the other hand, is a variant of [`Option`](https://doc.rust-lang.org/std/option/enum.Option.html), which is *not* returned by `canonicalize`.

PR #1174, opened 2022-10-06
Fix known surface duplication issues

hannobraun:
This builds upon the previous work done in #1168, #1170, #1172, and other pull requests, fixing all other known surface duplication issues and updating any surface-related validation code to use surface identity instead of surface equality, making this validation code much stricter and preventing possible bugs. This was made possible by advancements in #1021.

See the documentation added in this commit for a detailed explanation:
https://github.com/hannobraun/Fornjot/blob/87612f1e04be115528ba7dc929fef5bc442348e7/crates/fj-kernel/src/objects/mod.rs#L8-L71


hannobraun:
CI is failing because I overlooked another problem that I accidentally introduced. I'm looking into it.

hannobraun:
Figured it out. These damn bugs can be to subtle... good thing we have validation code!

PR #1199, opened 2022-10-10
Added Nix build and dev-shell support via Nix flakes

Philipp-M:
This PR adds support for Nix flakes (and the legacy nix build system via `flake-compat`).

Tested on NixOS, it may or not may run on other (linux/osx?) distributions.

I hacked this together to run and (likely) package Fornjot on NixOS, FWIW it may be helpful for other Nix users as well.

I can extend the `README` as well if it's helpful (Where would it fit best, in case?)

This can be run with (nix flakes enabled):

```bash
nix run github:Philipp-M/Fornjot/nix-flakes
```

or (without nix flakes enabled, in repo):

```bash
nix-build .
./result/bin/fj-app
```

Also supports a devshell environment, via `nix shell`

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,10 @@
+(import
+  (
+    let lock = builtins.fromJSON (builtins.readFile ./flake.lock); in
+    fetchTarball {
+      url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+      sha256 = lock.nodes.flake-compat.locked.narHash;
+    }
+  )
+  { src = ./.; }
+).defaultNix
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,110 @@
+{
+  "nodes": {
+    "crane": {
+      "inputs": {
+        "flake-compat": [
+          "flake-compat"
+        ],
+        "flake-utils": [
+          "flake-utils"
+        ],
+        "nixpkgs": [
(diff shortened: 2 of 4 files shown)


hannobraun (approved):
Thank you for this pull request, @Philipp-M!

I don't know enough about Nix or NixOS to review this properly, but I'm happy to trust your word on that it works and that it may be helpful to other Nix users. Happy to merge, but I'd like to make a few things clear:

- I don't use Nix or NixOS, so I won't notice if this breaks. People will have to open issues, if that happens.
- I don't have the skills to maintain this, nor the bandwidth and will to learn. If this is going to need updates going forward, those will have to be made by contributors.
- If this ends up broken, and no one steps up to fix it for a while, I might decide to remove it.

Does that sound reasonable?


> I can extend the `README` as well if it's helpful (Where would it fit best, in case?)

There's a section on installation: https://github.com/hannobraun/Fornjot#installation

I think it would be appropriate to add a note there.

---

I'm merging this now, as I'm happy with it as-is. Any further improvements (like an addition to the `README`) would be very welcome in a follow-up PR.

hannobraun:
Question: Would it be practical to move all those files into a subdirectory? Maybe `nix/` or `packages/nix/`?

If so, this would be another welcome addition that I'd be happy to merge.

hannobraun:
Thank you, @Philipp-M, that's great!

PR #1230, opened 2022-10-17
Add Windows support to Export Validator

hannobraun:
This fails the CI build. Export Validator builds fine, but I just can't get Windows to find the DLL file at runtime.

I don't know what else to try. Any help is appreciated!


hannobraun:
There's an open issue about Windows support for `export-validator` (https://github.com/hannobraun/Fornjot/issues/1231), which references this pull request. No need to keep it open.

PR #1233, opened 2022-10-18
Fix last known `SurfaceVertex` duplication issue

hannobraun:
Make use of the inclusion of `SurfaceVertex` in the centralized object storage (#1021) to make the cycle validation code stricter (by comparing based on [identity instead of equality](https://docs.rs/fj-kernel/0.20.0/fj_kernel/objects/index.html#object-identity-vs-object-equality)). This surfaced a few object duplication issues, some of which were already fixed. This takes care of the last known one.


hannobraun:
Ah, looks like I missed one. Working on it!

Issue #971, opened 2022-08-22
Print timestamp with each status update


hannobraun:
The Fornjot app has recently started displaying status updates, indicating that a model has been reloaded, or that an error occurred while trying to do just that.

If the model has previously been loaded successfully for a few times, it can be hard to judge whether a recent change has been registered. Unless it causes an error, the status update will look identical to the previous ones that were already there.

It would be great, if status updates showed a timestamp. Ideally, it would show how long ago that status update was, but as an initial implementation, just showing the time will be more than enough.

hannobraun:
The status messages are controlled by the [`StatusReport` struct in `fj-interop`](https://github.com/hannobraun/Fornjot/blob/2cb0c7a282dec550336f563860a52b2e3d0e5132/crates/fj-interop/src/status_report.rs). Adding the timestamp should be a relatively straight-forward addition to the `update_status` method.

Labeling https://github.com/hannobraun/Fornjot/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22, since this is a fairly small and isolated change.

PR #1256, opened 2022-10-23
Print timestamp with each status update

erenoku:
The timestamps are added with the format `[%H:%M:%S]` to the beginning of status updates.

Closes https://github.com/hannobraun/Fornjot/issues/971


hannobraun (approved):
Thank you, @erenoku, this is great!

PR #1266, opened 2022-10-24
Fix blurry egui

erenoku:
The egui context created here: https://github.com/hannobraun/Fornjot/blob/8c423093840ac7bb99b56533cb8b5804d43a3836/crates/fj-viewer/src/gui.rs#L52 for some reason didn't have the right pixels_per_point ratio so I explicitly set this ratio with the value I got from winit. This solved the blurriness but unfortunately the mouse input is still offseted.

### DIFF
--- a/crates/fj-viewer/src/graphics/renderer.rs
+++ b/crates/fj-viewer/src/graphics/renderer.rs
@@ -156,8 +156,8 @@ impl Renderer {
         })
     }
 
-    pub(crate) fn init_gui(&self) -> Gui {
-        Gui::new(&self.device, self.surface_config.format)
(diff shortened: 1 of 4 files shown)


hannobraun (approved):
Thank you, @erenoku, this is awesome! I had already searched for something like this, but somehow I've missed it.

Your discovery inspired me to search again, and it turns out that `egui-winit` has a similar method, and calling that with the right value fixes the input too! I've pushed some additional commits to do that.

While I was at it, I also moved the call to `set_pixels_per_point` to the update method. This won't make a difference in most cases, but it doesn't seem to hurt and seems more robust.

Thanks again! Always nice to get a second set of eyes on these kinds of things. It's so easy to miss things, when you're working by yourself most of the time.

PR #1286, opened 2022-10-28
Handle no model argument being passed

erenoku:
This PR adds support for starting fj-app with no arguments.

If no model argument is passed or no `fj.toml` file is present start the UI normally but display a warning and a button to open file picker in the middle of the screen. If the app is in export only mode display `no_model_error` and exit. The communication between `egui` and the main `event_loop` is handled by `crossbeam_channels` and the `Gui` struct stores a `GuiState`  to store whether there is a model opened.


erenoku:
Turns out the libraries `gtk-3` or `xdg-portal` packages are needed to compile the crate [RFD](https://lib.rs/crates/rfd). That's why the CI jobs are failing. I think we can use the [CI file of that repository](https://github.com/PolyMeilex/rfd/blob/76b8e30a63fadc130838f0231530726e04fbd356/.github/workflows/rust.yml) to make our CI tests work.

hannobraun:
Thank you, @erenoku! I don't have time to review right now, but I'll take a look soon. Tomorrow, if I can, or next week, if not.

crates/fj-viewer/src/gui.rs
```diff
@@ -14,22 +14,47 @@
...
 use fj_math::{Aabb, Scalar};
 
-use crate::graphics::DrawConfig;
+use crate::{file_dialog::show_file_dialog, graphics::DrawConfig};
+
+/// Event that are passed between the event_loop and gui
+pub enum GuiEvent {
+    AskModel,
+    LoadModel(PathBuf),
+}
```
hannobraun, line 31: Maybe I'm missing something, but it looks like `AskModel` is only sent to the GUI from the main loop, while `LoadModel` is only sent by the GUI to the main loop. Both `match` statements that handle `GuiEvent` only handle one variant, while ignoring the other.

It seems to me like `GuiEvent` shouldn't exist, and `AskModel`/`LoadModel` should be separate structs instead.

crates/fj-window/src/run.rs
```diff
@@ -31,14 +31,25 @@ pub fn run(
     shape_processor: ShapeProcessor,
     invert_zoom: bool,
 ) -> Result<(), Error> {
+    let (send_gui, gui_event_rx) = crossbeam_channel::bounded::<GuiEvent>(5);
+
+    let (gui_event_tx, recv_gui) = crossbeam_channel::bounded::<GuiEvent>(5);
```
hannobraun, line 35: Those bounded channels with capacity `5` look a bit fishy to me. Why 5? Is there some chance that these buffers can fill up and cause a deadlock?

I usually create bounded channels with capacity `0`, because if the channel *can* theoretically cause a deadlock, it usually *will* cause a deadlock if it has no buffer! If it doesn't cause a deadlock and works otherwise, it should be fine to leave it at capacity `0`, unless there are specific reasons not to (like performance).

crates/fj-app/src/main.rs
```diff
@@ -66,3 +67,11 @@ fn main() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+fn no_model_error() -> anyhow::Error {
+    anyhow!(
+        "You must specify a model to start Fornjot.\n\
+        - Pass a model as a command-line argument. See `fj-app --help`.\n\
+        - Specify a default model in the configuration file."
+    )
```
hannobraun, line 76: This error message isn't quite true anymore. It only appears in the context of exporting.

I think it would be nicer to update it, to specifically say you need a model *when exporting*. But this is optional, and I'd happily merge the pull request with the error message staying like this.

crates/fj-viewer/Cargo.toml
```diff
@@ -12,11 +12,13 @@ categories.workspace = true
 
 [dependencies]
 bytemuck = "1.12.1"
+crossbeam-channel = "0.5.6"
 egui = "0.19.0"
 egui-wgpu = "0.19.0"
 fj-interop.workspace = true
 fj-math.workspace = true
 raw-window-handle = "0.4.3"
+rfd = "0.10.0"
```
hannobraun, line 21: It looks like by default, RFD uses GTK3 on Linux: https://github.com/PolyMeilex/rfd/blob/7a0ab7a0ec756b8cb8e30fba242ca86c372ee996/Cargo.toml#L14

This works fine on my machine, but I'm using Gnome, which is GTK-based. I don't think it will work everywhere. Based on the RFD documentation, it seems like [XDG Desktop Portal might be the better choice](https://docs.rs/rfd/0.10.0/rfd/index.html#xdg-desktop-portal-backend). It [supports fewer features](https://docs.rs/rfd/0.10.0/rfd/index.html#state), but I don't think we're using any of those anyway?

We should probably disable default features here, and enable the [`xdg-portal`](https://github.com/PolyMeilex/rfd/blob/7a0ab7a0ec756b8cb8e30fba242ca86c372ee996/Cargo.toml#L17) feature.

hannobraun (changes requested):
Thank you, @erenoku! This is great!

I've taken a close look at the code and left some comments. In addition, there are a few more problems that need to be addressed before this is ready to merge:

- The CI build needs fixing, as you note. It looks like the CI file you linked doesn't need to install anything for XDG Desktop Portal, so the build might fix itself, if we use that instead (I've also left a comment on that topic).
- It's not up-to-date with the `main` branch and needs a rebase. There are no conflicts, so I could do this automatically when merging, but I'd like to see how this looks together with https://github.com/hannobraun/Fornjot/pull/1275, and if maybe a similar fix is required for the new UI element.
- There are some compiler and Clippy warnings that need fixing. Those would fail the CI build, once the other CI issue is addressed.
- There's a panic when quitting the application:
  ```
  thread '<unnamed>' panicked at 'Expected channel to never disconnect: RecvError', crates/fj-host/src/evaluator.rs:22:18
   ```
  I'm actually not sure if this is new, or if it was there before. In any case, it would be nice to get it fixed, but it's not critical.

crates/fj-window/src/run.rs
erenoku: I changed the capacity to be 1. I am not sure why but a capacity of zero caused the GUI not to be initialized. I think the reason is that there is a single event loop and both the send and receive operations are ran on the same thread so a buffer size of 1 is necessary.

erenoku:
Thanks for the feedback.
I think all the issues are fixed now but I could not reproduce the panic when quitting the application. If it consistently happens on your machine could you upload the backtrace.

crates/fj-viewer/src/gui.rs
```diff
@@ -24,12 +24,6 @@ use fj_math::{Aabb, Scalar};
 
 use crate::graphics::DrawConfig;
 
-/// Event that are passed between the event_loop and gui
-pub enum GuiEvent {
-    AskModel,
-    LoadModel(PathBuf),
-}
```
hannobraun: Removing `GuiEvent` outright is perfectly fine, but just for the record, I think it would have been slightly nicer to keep a `struct AskModel;` and a `struct LoadModel(PathBuf);` around, to make those `Sender`/`Receiver` types below a bit more descriptive (`Sender<AskModel>` instead of `Sender<()>`, for example).

But this is only a minor thing! Just mentioning it for future reference.

crates/fj-viewer/src/gui.rs
```diff
@@ -348,9 +353,13 @@ impl Gui {
...
 fn show_file_dialog() -> Option<PathBuf> {
-    FileDialog::new()
+    #[cfg(not(target_arch = "wasm32"))]
+    return FileDialog::new()
         .set_directory(current_dir().unwrap_or_else(|_| PathBuf::from("/")))
-        .pick_folder()
+        .pick_folder();
+
+    #[cfg(target_arch = "wasm32")]
+    todo!("Picking folders does not work on wasm32")
```
hannobraun, line 362: This (and the other `cfg`s) aren't great, but I can't think of a better way to do it right now. Maybe the file picking feature should move out of `fj-viewer` and into `fj-window` or `fj-app`? No idea. We'll figure it out at some point!

crates/fj-host/src/host.rs
```diff
@@ -16,7 +16,10 @@ impl Host {
     pub fn from_model(model: Model) -> Result<Self, Error> {
         let watch_path = model.watch_path();
         let evaluator = Evaluator::from_model(model);
-        let _watcher = Watcher::watch_model(&watch_path, &evaluator)?;
+        let _watcher = match Watcher::watch_model(&watch_path, &evaluator) {
+            Ok(_watcher) => _watcher,
+            Err(e) => return Err(e),
+        };
```
hannobraun, line 22: What is this change about? It doesn't seem to make a difference, except make the code longer?

crates/fj-window/src/run.rs
hannobraun: Yeah, it's probably something like that :+1:

hannobraun:
Hmm, looks like GitHub is being weird and sent my review comments before I actually wrote the main review itself. Please stand by, I'll finish that shortly.

hannobraun (approved):
Thank you for the changes, @erenoku!

I've left some comments, but all of those are about minor things. I'll merge this right now. Feel free to create a follow-up PR with additional changes, if you want, or don't :smile: 

> I think all the issues are fixed now but I could not reproduce the panic when quitting the application. If it consistently happens on your machine could you upload the backtrace.

There is no backtrace, beyond the file/line from the panic message, as that panic happens in the root scope of one of the threads.

If you're not seeing it, don't worry. It's not critical anyway, and it'll be much easier for me to debug, since I can reproduce it consistently. I've made a note about that, and will take a look next week, mostly likely.

crates/fj-host/src/host.rs
erenoku: You're right. This was needed when `from_model` function was returning an `Option<Result<Self, Error>>` but I forgot to change this to the original when converting it back to returning a `Result<Self, Error>`.

hannobraun: Ah, got it. Thanks for the follow-up PR!

PR #1300, opened 2022-11-02
Warn on full version mismatch of host and model

### LINKED ISSUE #1280: Print warning, if full version of app and model don't match
Currently, if the version of the Fornjot app doesn't match the version of Fornjot that a model uses, this will result in an error when the model is loaded. This works well enough for released versions, but the it might lead to confusion when in-development versions are involved. This might be the case, if a user tries to load an example model from this repository using a released version of the Fornjot app.

The version is determined at build time in [the `fj` crate's `build.rs`](https://github.com/hannobraun/Fornjot/blob/147f7e8cbe12875c5705296eb4c463c8735f210a/crates/fj/build.rs). From there, it is [included in the `fj` crate](https://github.com/hannobraun/Fornjot/blob/147f7e8cbe12875c5705296eb4c463c8735f210a/crates/fj/src/version.rs#L5-L20), and the `static`s defined there are then [used at model-load-time in `fj-host`](https://github.com/hannobraun/Fornjot/blob/147f7e8cbe12875c5705296eb4c463c8735f210a/crates/fj-host/src/model.rs#L107-L121) to check for a mismatch and potentially generate the error. Two different versions are determined and included in `fj`: The package version (something like `1.2.3`), and a "full" version which includes more information, like the Git commit hash. Only the package version is used for the version check.

It would be too restrictive to use the full version as the basis for the version mismatch check, but since a mismatch there *might* cause mysterious problems, it's probably best to print a warning, explaining the situation. I'm not sure what the ideal solution is, but for a start, it would probably be enough to log a warning-level log message to stderr.

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this only requires surface-level knowledge of Fornjot, and I've laid out all the moving parts here.

zthompson47:
Fixes #1280

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1170,6 +1170,7 @@ dependencies = [
  "libloading",
  "notify",
  "thiserror",
+ "tracing",
 ]
 
 [[package]]
--- a/crates/fj-host/Cargo.toml
+++ b/crates/fj-host/Cargo.toml
@@ -18,3 +18,4 @@ fj.workspace = true
 libloading = "0.7.2"
 notify = "5.0.0"
 thiserror = "1.0.35"
+tracing = "0.1.37"
--- a/crates/fj-host/src/model.rs
+++ b/crates/fj-host/src/model.rs
@@ -6,6 +6,7 @@ use std::{
 };
 
 use fj::{abi, version::RawVersion};
+use tracing::warn;
 
 use crate::{platform::HostPlatform, Parameters};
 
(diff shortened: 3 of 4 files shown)


hannobraun (approved):
Thank you, @zthompson47, this looks great!

I didn't try to test this (running low on time for today), but since [your own testing approach](https://github.com/hannobraun/Fornjot/issues/1280#issuecomment-1300310432) looks reasonable, as does your code, I'm happy to merge!

PR #1323, opened 2022-11-07
Turn the "--models" CLI option into an argument

### LINKED ISSUE #1272: Remove `-m`/`--model`
`fj-app` can be provided with the `-m`/`--model` argument, to specify which model to load. And in fact, this is the only command-line argument that isn't really optional (strictly speaking it is, but only if a default model is specified in the configuration file).

It would be simpler to just do away with `-m`/`--model` and use a positional argument instead. So instead of this:
```
fj-app -m model
```

We'd have this:
```
fj-app model
```

I think this would be more natural. In fact, I forget the `-m`/`--model` all the time.

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this only requires surface-level knowledge of Fornjot. This probably requires changes only in [`fj-app`'s `args.rs`](https://github.com/hannobraun/Fornjot/blob/main/crates/fj-app/src/args.rs).

kopackiw:
This PR removes the `--model` flag in the CLI in favour of an optional `model` argument.
Tested locally on custom inputs & examples from the documentation - no regression found.

Closes #1272

### DIFF
--- a/fj.toml
+++ b/fj.toml
@@ -1,9 +1,9 @@
-# The default path that models are loaded from. If the `--model` argument is a
+# The default path that models are loaded from. If the `model` argument is a
 # relative path, it is assumed to be relative to `default_path`.
 default_path = "models"
 
-# The default models that is loaded, if none is specified. If this is a relative
-# path, it should be relative to `default_path`.
+# The default models that is loaded, if none is specified in a CLI.
+# If this is a relative path, it should be relative to `default_path`.
 default_model = "test"
 
 # Indicate whether to invert the zoom direction. Can be used to override the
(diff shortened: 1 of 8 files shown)


fj.toml
```diff
@@ -1,9 +1,9 @@
-# The default path that models are loaded from. If the `--model` argument is a
+# The default path that models are loaded from. If the `model` argument is a
```
kopackiw, line 1: One question to consider: should we uppercase any names of CLI arguments used in the documentation (as it is formatted now in `help`)?

Screenshot 2022-11-07 at 23 54 35

hannobraun: Good point! Yes, I think we should.

(Still going to merge this as-is though. This is not critical and can always be done in a follow-up PR.)

hannobraun (approved):
Thank you @kopackiw, this is great!

Issue #883, opened 2022-07-28
All binaries built by CD workflow are labeled as release binaries


hannobraun:
As of #868, we now have a `--version` argument that displays the version, and makes a distinction between release and development versions. Unfortunately all binaries built by the CD workflow are labeled as release binaries for the purpose of the `--version` flag. This leads those binaries that aren't promoted to release binaries to display a wrong version.

The way it works, is that on every push to `main`, the CD workflow is started, and builds binaries:
https://github.com/hannobraun/Fornjot/blob/016d5d2b2ff2dd33888c1b25d38d47c1f2c4749d/.github/workflows/cd.yml#L23-L77

The flag that marks the binaries as official release binaries is wrongly set for that whole workflow:
https://github.com/hannobraun/Fornjot/blob/016d5d2b2ff2dd33888c1b25d38d47c1f2c4749d/.github/workflows/cd.yml#L14-L16

After all binaries are built, the release job starts, and will publish a release (using those binaries), if that's what is supposed to happen. Whether it's supposed to happen, is determined by the `Operator | Deduce` step:
https://github.com/hannobraun/Fornjot/blob/016d5d2b2ff2dd33888c1b25d38d47c1f2c4749d/.github/workflows/cd.yml#L101-L109

At this point, the binaries have already been built.

I see two ways to address this:

1. Move the `Operator | Deduce` step to a separate job that runs before the binaries are built. Only set `FJ_OFFICIAL_RELEASE`, if it detects that a release should happen.
2. Don't build binaries at all, unless a release is detected. We have weekly releases now, so maybe we don't need the per-build binaries anymore?

I think I'd prefer solution 1, but I'd accept solution 2.

Labeling as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, since this only requires knowledge of GitHub Actions to address, not any deeper understanding of Fornjot in particular.

hannobraun:
@hendrikmaus has left a [comment in #850](https://github.com/hannobraun/Fornjot/issues/850#issuecomment-1198004068) with a proposal on how to approach a solution to this issue.

kopackiw:
I'd like to work on that issue, but need to clarify some things first.

We have two flags:
1. `FJ_OFFICIAL_RELEASE` (currently hardcoded to `1`) which tells whether the release is official (e.g. production one) or not (e.g. development one).
[...]

Issue #1270, opened 2022-10-26
Fix CD warning


hannobraun:
The continuous deployment build currently emits two warnings: https://github.com/hannobraun/Fornjot/actions/runs/3319079604

I believe the one about macOS is not relevant for us, but this one is:

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

According to the linked blog post, **this will stop working 2023-05-31**.

I believe this piece of code is responsible for the warning:
https://github.com/hannobraun/Fornjot/blob/63a7518c29956c9be2bd49b36c868b41f73b2150/tools/release-operator/src/github.rs#L71-L75

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this issue doesn't require any knowledge about Fornjot to address.

kopackiw:
Happy to work on that later today!

PR #1324, opened 2022-11-08
Unify release flags and correctly mark dev build releases

kopackiw:
❗NOT TESTED❗

This PR unifies the `FJ_OFFICIAL_RELEASE` and `release-detected` flags that carry the same meaning.

To properly read the `release-detected` as env, work in https://github.com/hannobraun/Fornjot/issues/1270 is required I think.

Closes https://github.com/hannobraun/Fornjot/issues/883

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -40,6 +40,7 @@ jobs:
 
   binaries:
     name: Binaries
+    needs: calculate-release-flags
     strategy:
       matrix:
         include:
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -20,7 +20,7 @@ jobs:
     name: Calculate release flags
     runs-on: ubuntu-latest
     outputs:
-      release-detected: ${{ steps.step1.outputs.release-detected }}
+      release-detected: ${{ steps.release.outputs.release-detected }}
 
     steps:
       - name: Checkout
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -19,6 +19,9 @@ jobs:
   calculate-release-flags:
     name: Calculate release flags
     runs-on: ubuntu-latest
+    outputs:
+      release-detected: ${{ steps.step1.outputs.release-detected }}
+
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -52,7 +52,7 @@ jobs:
           - { target: aarch64-apple-darwin, os: macOS-latest }
           - { target: x86_64-pc-windows-msvc, os: windows-latest }
 
-    runs-on: ${{matrix.os}}
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -91,7 +91,7 @@ jobs:
 
   release:
     name: Release
-    if: ${{ steps.calculate-release-flags.outputs.release-detected == 'true' }}
+    if: ${{ needs.calculate-release-flags.outputs.release-detected == 'true' }}
     needs: binaries
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -68,6 +68,8 @@ jobs:
       - name: Binaries | Compile
         run: cargo build --release --target ${{ matrix.target }}
         env:
+          # This lets our app know it's an "official" release. Otherwise we would get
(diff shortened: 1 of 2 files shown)


crates/fj/build.rs
```diff
@@ -23,8 +23,8 @@ impl Version {
         let commit = git_description();
 
         let official_release =
-            std::env::var("FJ_OFFICIAL_RELEASE").as_deref() == Ok("1");
-        println!("cargo:rerun-if-env-changed=FJ_OFFICIAL_RELEASE");
+            std::env::var("release-detected").as_deref() == Ok("true");
```
kopackiw, line 26: The name can be changed, as for me it is a bit misleading. I propose e.g. `is-official-release`.

hendrikmaus: The release operator doesn't set any environment variables, it sets github actions outputs.
So if you want the `build.rs` to have access to the data, you need to export it in the step that will call `cargo build`.
At a glance, I'd say say this needs to happen here https://github.com/hannobraun/Fornjot/pull/1324/files#diff-ea3ea8c9932adc7ba8161ceda844fedd43b006848ef1140c050cbd7ea0788a18R65

Environment variables, as a best practice, are always written in all caps. So I would propose to call it `RELEASE_DETECTED` then.

On the aforementioned step - the one that runs `cargo build` - we could then populate the variable like so:

```yaml
- name: Some step
  env:
    RELEASE_DETECTED: needs.calculate-release-flags.outputs.release-detected
  # ...
```

hendrikmaus (changes requested):

.github/workflows/cd.yml
```diff
@@ -69,6 +88,7 @@ jobs:
 
   release:
     name: Release
```
hendrikmaus, line 90:
```suggestion
    name: Release
    needs: calculate-release-flags
```

.github/workflows/cd.yml
```diff
@@ -69,6 +88,7 @@ jobs:
...
+    if: ${{ steps.calculate-release-flags.outputs.release-detected == 'true' }}
     needs: binaries
```
hendrikmaus, line 92:
```suggestion
    needs:
      - binaries
      - needs: calculate-release-flags
```

crates/fj/build.rs
hannobraun: I'm open to renaming this output/variable, although I'm not sure about `is-official-release`. There's no such thing as an unofficial release, so what exactly does that mean? (Yes, I realize the previous environment variable did the same thing.) I agree with @hendrikmaus' point about capitalization.

Personally, I'd be fine with `RELEASE_DETECTED`, but that's just bikeshedding. I'd merge this pull request with any semi-reasonable naming choice. We can always improve things in follow-up PRs.

hannobraun (commented):
Thank you for this pull request, @kopackiw!

This looks good to me (modulo the problem with the output/variable that @hendrikmaus pointed out). Regarding testing, I assume it would be a pain to set up a test outside of this repository, so let's just merge it and see how it goes, once @hendrikmaus runs out of problems to point out :smile:

.github/workflows/cd.yml
```diff
@@ -11,17 +11,36 @@ env:
...
-  # This lets our app know it's an "official" release. Otherwise we would get
-  # a version number like "fj-app 0.8.0 (8cb928bb, unreleased)"
-  FJ_OFFICIAL_RELEASE: 1
-
 defaults:
   run:
     shell: bash
 
 jobs:
+  calculate-release-flags:
```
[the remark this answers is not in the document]
kopackiw: Good catch, fixed 👍🏻

.github/workflows/cd.yml
```diff
@@ -69,6 +88,7 @@ jobs:
 
   release:
     name: Release
+    if: ${{ steps.calculate-release-flags.outputs.release-detected == 'true' }}
```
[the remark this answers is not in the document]
kopackiw: Fixed, thank you!

.github/workflows/cd.yml
kopackiw: I wrote that way because the `calculate-release-flags` is an indirect dependency via the `binaries`. But it won't work with an if expression above: `if: ${{ needs.calculate-release-flags.outputs.release-detected == 'true' }}`, so fixed as well!

hannobraun (approved):
Thank you for the changes, @kopackiw!

As far as I'm concerned, this is good to go, but I'd like to give @hendrikmaus the chance to take another look. I'll merge as soon as he gives the go-ahead!

hendrikmaus (approved):
Thanks @hannobraun, I think we're good to go.

hannobraun:
Thanks, @hendrikmaus! Merging.

PR #1335, opened 2022-11-10
fix: wrap assignment with quotes

kopackiw:
After merging [the PR that remove deprecated set-input](https://github.com/hannobraun/Fornjot/commit/6d7a48ec5115f35a025fe917227b82f36cfb911f), I saw that `RELEASE_DETECTED` [flag](https://github.com/hannobraun/Fornjot/actions/runs/3432020181/jobs/5720842261#step:5:9) is set to an empty string (it should be `"false"` or `"true"`). Thus, weekly official release might not work.

I think that problem might lie in quoting. If the flag on CD after merge still be an empty string, please @hannobraun revert this and the former PR to not disrupt your release pattern. Sorry for trouble 🙏🏻

### DIFF
--- a/tools/release-operator/src/github.rs
+++ b/tools/release-operator/src/github.rs
@@ -71,6 +71,6 @@ impl Actions {
     // Set an "output" in GitHub Actions
     pub fn set_output(key: Outputs, value: &str) {
         log::debug!("setting output name={key} value={value}");
-        println!("{key}={value} >> $GITHUB_OUTPUT");
+        println!("\"{key}={value}\" >> $GITHUB_OUTPUT");
     }
 }


hannobraun (approved):
Thank you for this pull request, @kopackiw! Good catch on the variable being empty. I had a look myself, but missed this.

> Sorry for trouble 🙏🏻

Please, no need to apologize! I'm very thankful that you're taking care of this. It might only be a small thing, but it's one of a million little things that I don't have time for, so again, thank you!

The risk of disrupting the release process is acceptable. As I said earlier somewhere, I think it would be a big pain to set up proper testing for this. It's just unreasonable to expect this, when we have a weekly release process that can serve as the perfect testing ground.

If any issues are still present after this, they'll get ironed out over the next few weeks. No big deal!

PR #1342, opened 2022-11-13
Run export-validator on Windows

### LINKED ISSUE #1231: `export-validator` does not support Windows
[`export-validator`](https://github.com/hannobraun/Fornjot/tree/main/tools/export-validator) is an application in this repository that uses the Fornjot app to export all models in the [`models/`](https://github.com/hannobraun/Fornjot/tree/main/models) directory to 3MF files, and to validate the correctness of those 3MF files (using [`lib3mf`](https://github.com/3MFConsortium/lib3mf)). This is extremely valuable, as it catches a lot of bugs as part of our extended test suite. Specifically, `export-validator` is run on every CI build, and can be run locally via [`just build`](https://github.com/hannobraun/Fornjot/blob/main/justfile).

`lib3mf` is a C++ library, which makes the whole setup a bit problematic. I figured it would be easiest to just use the pre-compiled binaries from their SDK. This works for Linux and macOS, but I failed to get this to work for Windows, meaning `export-validator` doesn't run on Windows at all. This is a problem in two ways:

- Windows developers can't use `just build` or `cargo run -p export-validator` locally, which limits their ability to test their changes without submitting a pull request and running a CI build.
- The CI build has reduced test coverage on Windows, which might allow platform-specific bugs to slip through.

For these reasons, it would be great if `export-validator` supported Windows too. In principle, this would involve the following steps:

1. Extract the pre-compiled Windows libraries from the [lib3mf SDK](https://github.com/3MFConsortium/lib3mf/releases/tag/v2.2.0).
2. Add those files to [the `libs/` directory](https://github.com/hannobraun/Fornjot/tree/main/tools/export-validator/lib3mf/libs) under the correct names, so they can be found when linking.
3. Adapt [`export-validator`'s `build.rs`](https://github.com/hannobraun/Fornjot/blob/main/tools/export-validator/build.rs) so everything works.
4. Enable `export-validator` in the Windows CI build, by removing [the conditional that prevents it from running](https://github.com/hannobraun/Fornjot/blob/main/.github/workflows/ci.yml#L134-L136).

I tried to do this, but I just couldn't get it to work. I managed to link to the DLL, but I couldn't make it so that the DLL would then be found at load-time (i.e. when `export-validator` runs). Here is my aborted attempt: https://github.com/hannobraun/Fornjot/pull/1230

I would really appreciate some help with this. I don't have Windows, don't know a whole lot about it, and testing this only via CI builds is a pain in any case. Labeling as https://github.com/hannobraun/Fornjot/labels/help%20wanted.

MartinKavik:
Add `export-validator` Windows support.

Fixes #1231 

Tested with `just build` on **Windows 10** and **Kubuntu 22.04** local machines.

---

_Note:_ 
To make `just build` work on Windows I had to install [`Visual Studio` with `clang`](https://www.wikihow.com/Install-Clang-on-Windows) and then [`MinGW` with `ar`](https://winlibs.com/) and add both tools to `PATH` env var to make `cross-compiler` work.
On Linux, I had to install `libfontconfig-dev` and `clang`.


hannobraun (approved):
Thank you, @MartinKavik! Looks reasonable, and the CI build is green. Good to merge :+1: 

> _Note:_
> To make `just build` work on Windows I had to install [`Visual Studio` with `clang`](https://www.wikihow.com/Install-Clang-on-Windows) and then [`MinGW` with `ar`](https://winlibs.com/) and add both tools to `PATH` env var to make `cross-compiler` work.
> On Linux, I had to install `libfontconfig-dev` and `clang`.

Something like this should probably go into `CONTRIBUTING.md`. I've made a note about that, and plan to take care it next week (either by doing that, or by opening an issue).

PR #1344, opened 2022-11-14
Command to create a new model

### LINKED ISSUE #1271: Add command to generate model
When someone wants to try Fornjot, there's not a great way to see something quickly, unless they're cloning the Git repository and use the example models in there. For that reason, and also to make it easier to get started in general, it would be great if the Fornjot app could generate an example model.

Eventually, this should be possible from within the app, but for now, a command-line argument would suffice. Maybe something like this:
```
fj-app model new <name>
```

This would generate a model in the `<name>/` directory within the current one. The generated model should be something simple. [`cuboid`](https://github.com/hannobraun/Fornjot/tree/main/models/cuboid) or [`spacer`](https://github.com/hannobraun/Fornjot/tree/main/models/spacer) could be good candidates. Maybe even [`star`](https://github.com/hannobraun/Fornjot/tree/main/models/star). It doesn't actually matter much. Once the infrastructure is in place, it should be easy to change which model is being generated.

Ideally, the generated model should be regularly tested, so it doesn't become stale. Maybe including one of the [example models](https://github.com/hannobraun/Fornjot/tree/main/models) directly in the binary using something like [`include_dir`](https://crates.io/crates/include_dir) would be an option. Those are part of the CI build, so they should always be in sync with the app.

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue. This will require some figuring out and design decisions, but as far as Fornjot is concerned, it only requires surface knowledge. The place to add the command-line argument is [`args.rs` in `fj-app`](https://github.com/hannobraun/Fornjot/blob/9dcd24a978b1db893a15de9306e60c83f9f1cfc1/crates/fj-app/src/args.rs), and any of the code for generating the model can just go into a new module within `fj-app`.

MartinKavik:
Resolves #1271 

Adds the Clap arg `new` to create a new _model_ crate, ex.: 
```
cargo run -- --new my_new_model
```
It creates a new folder `my_new_model` with the package/crate with the same name. The crate is based on the official example `models/star`.

---

Tested on Windows 10 and Kubuntu with `cargo run`.

The code is based on MoonZoon CLI tool [mzoon](https://github.com/MoonZoon/MoonZoon/tree/main/crates/mzoon) with removed `async` support. 

**Logic:** The example `models/star` (aka _template_) is packed to a Tar archive and included in `fb-app` as bytes. Ignored files (hidden, gitignored) and some extra ones are not archived. When the arg `--new` is processed, the archive is unpacked and files are postprocessed (for instance some text files are modified).

---

**Notes:**
- I didn't want to change Clap logic too much but I agree that something like `fb-app model new ` could be a bit better than `--new`. 
- Postprocessing currently replaces the relative path to `fj` to its `VERSION_PKG` in the new model package. We can add a new flag to don't do that so it's easier to test created packages. It looks like [this in `mzoon`](https://github.com/MoonZoon/MoonZoon/blob/6b7b3f12cb2ef6fd4f067ea7079c2be8992394ae/crates/mzoon/src/main.rs#L21-L28)
- We can consider to make the commands async in the future - I can image some commands like _"export_all_models"_ will need parallelization.  The structure might look like [this](https://github.com/MoonZoon/MoonZoon/blob/6b7b3f12cb2ef6fd4f067ea7079c2be8992394ae/crates/mzoon/src/main.rs#L96-L116). 

---

P.S. The design & implementation of this feature is quite opinionated. Feel free to close the PR or create a new PR inspired by it. Don't hesitate to copy code from `mzoon` if it helps.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -347,6 +347,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0d8c1fef690941d3e7788d328517591fecc684c084084702d6ff1641e993699a"
 
 [[package]]
+name = "bstr"
+version = "0.2.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba3569f383e8f1598449f1a423e72e99569137b47740b1da11ef19af3d5c3223"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1145,7 +1154,9 @@ dependencies = [
  "fj-operations",
  "fj-viewer",
  "fj-window",
+ "ignore",
  "serde",
+ "tar",
  "tracing-subscriber",
 ]
 
@@ -1499,6 +1510,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "22030e2c5a68ec659fde1e949a745124b48e6fa8b045b7ed5bd1fe4ccc5c4e5d"
 
 [[package]]
+name = "globset"
+version = "0.4.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0a1e17342619edbc21a964c2afbeb6c820c6a2560032872f397bb97ea127bd0a"
+dependencies = [
+ "aho-corasick",
+ "bstr",
+ "fnv",
+ "log",
+ "regex",
+]
+
+[[package]]
 name = "glow"
 version = "0.11.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


hannobraun (approved):
Thank you, @MartinKavik, this looks great!


> I didn't want to change Clap logic too much but I agree that something like `fb-app model new ` could be a bit better than `--new`.

That's fine. We can always look into improvements later. In addition, there has been a change since I opened #1271 (models are loaded with `fj-app my-model` instead of `fj-app --model my-model`), so it might not be practical to add a subcommand-style interface anymore.

Doesn't matter now. We can rethink all of this later.


> * Postprocessing currently replaces the relative path to `fj` to its `VERSION_PKG` in the new model package. We can add a new flag to don't do that so it's easier to test created packages. It looks like [this in `mzoon`](https://github.com/MoonZoon/MoonZoon/blob/6b7b3f12cb2ef6fd4f067ea7079c2be8992394ae/crates/mzoon/src/main.rs#L21-L28)
> * We can consider to make the commands async in the future - I can image some commands like _"export_all_models"_ will need parallelization.  The structure might look like [this](https://github.com/MoonZoon/MoonZoon/blob/6b7b3f12cb2ef6fd4f067ea7079c2be8992394ae/crates/mzoon/src/main.rs#L96-L116).

Sounds reasonable :+1: 


> P.S. The design & implementation of this feature is quite opinionated. Feel free to close the PR or create a new PR inspired by it.

I think it looks great, although there might be alternatives (like [`include_dir`](https://crates.io/crates/include_dir)). I'd rather merge this now and improve upon it later, than leaving this open while trying to come up with some perfect solution.


> Don't hesitate to copy code from `mzoon` if it helps.

Thank you!

hannobraun:
It looks like this has broken `cargo install`: https://github.com/hannobraun/Fornjot/issues/1356

PR #1435, opened 2022-12-09
Fail on error parsing RUST_LOG

### LINKED ISSUE #1434: Fail on RUST_LOG parse error
At the top of `crates/fj-app/src/main.rs` there's this comment:
```
// It would be better to fail, if `RUST_LOG` is erroneous, but I don't know
// how to distinguish between that and the "not defined" case.
```
I figured out how to detect if RUST_LOG is not defined and came up with a PR that I'll attach to this issue.

My solution terminates the program with an error message if there is any type of parse error.  That includes not just [parsing errors in `tracing_subscriber`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.FromEnvError.html), but also the `NotUnicode` variant of [`std::env::VarError`](https://doc.rust-lang.org/stable/std/env/enum.VarError.html).  It succeeds only if the parsing is correct or if RUST_LOG is not defined (the latter case defaults to `WARN`).  Hopefully that's what you had in mind, but I can always revise.

I didn't write a test for this change, but I'd be happy to work on that if you think it's warranted.  I feel like it's something to just manually check and assume it will continue working.  That said, I've only checked the obvious cases, and e.g. haven't tested a non-unicode RUST_LOG.  I guess I'm just relying on the compiler guarantees.  Please advise if my thinking is off..

zthompson47:
Default to WARN log level if RUST_LOG is not defined or else terminate the program with an error message. Fixes #1434

### DIFF
--- a/crates/fj-app/src/main.rs
+++ b/crates/fj-app/src/main.rs
@@ -16,6 +16,8 @@ mod args;
 mod config;
 mod path;
 
+use std::{env, error::Error};
+
 use anyhow::{anyhow, Context};
 use fj_export::export;
 use fj_host::Parameters;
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @zthompson47! This looks great!

PR #1476, opened 2023-01-03
Improve model loading messages (long-lived background thread)

### LINKED ISSUE #1327: Shape processing is blocking event loop
When the Fornjot app loads a model, there are multiple steps to that:

1. The app determines that it needs to load the model, either because it detected a change in the model, or because this is the initial load.
2. The app starts compiling the model.
3. Once the model has been compiled, the app starts processing the model. This means it is evaluated, kernel data structures are built, and a triangle mesh (for rendering) is created.

This last step happens within the main event loop, which blocks the UI while that is going on. See this screenshot:
Screenshot from 2022-11-09 10-53-53

Those last two status messages have a clear difference in their timestamp, yet they show up in the UI at the same instant. The UI can't show the first one any earlier, because the main event loop is blocked.

The main event loop is [located in `fj-window`](https://github.com/hannobraun/Fornjot/blob/fed6533f54db8d76d05cd253e824367fce83be46/crates/fj-window/src/run.rs#L51) and uses [winit](https://crates.io/crates/winit). Winit has [a mechanism for passing outside events into the event loop](https://docs.rs/winit/0.27.5/winit/event_loop/struct.EventLoop.html#method.create_proxy), and I believe we should use that to do any non-UI work outside of the event loop and just pass the results in. I already looked into that, but unfortunately it is not a trivial change. The problem is, that the code operating outside of the event loop would need access to [`Host`](https://docs.rs/fj-host/0.23.0/fj_host/struct.Host.html), but the [`host` variable](https://github.com/hannobraun/Fornjot/blob/fed6533f54db8d76d05cd253e824367fce83be46/crates/fj-window/src/run.rs#L43) is mutable and [updated from within the event loop](https://github.com/hannobraun/Fornjot/blob/fed6533f54db8d76d05cd253e824367fce83be46/crates/fj-window/src/run.rs#L210), based on UI events.

We need to rethink the interaction between the event loop and `Host` to make this work. Either by making `Host` smarter, so it can handle no model being available and can react to a model being loaded, or by creating some kind of proxy between the event loop and `Host` that handles that.

### LINKED ISSUE #1431: Improve initial model loading message
Here's the status message displayed when first loading a model after starting the app:
Screenshot from 2022-12-09 11-15-17

It says "change in model detected", even though that isn't the case here. The model was loaded unconditionally, for the first time. It just reuses the same mechanism that is also used for reloading the model after a change was detected, hence the status message. It would be nicer, if an appropriate status message was displayed each time, that explained what actually happened.

The code that governs this is in `fj-host`. [`Watcher`](https://github.com/hannobraun/Fornjot/blob/39fadcd9fe5effd839bf99f64d88efe672b975ea/crates/fj-host/src/watcher.rs) watches the model for changes. When it detects a change, it triggers [`Evaluator`](https://github.com/hannobraun/Fornjot/blob/39fadcd9fe5effd839bf99f64d88efe672b975ea/crates/fj-host/src/evaluator.rs). Both of them are wrapped by [`Host`](https://github.com/hannobraun/Fornjot/blob/39fadcd9fe5effd839bf99f64d88efe672b975ea/crates/fj-host/src/host.rs), which is then [used in `fj-window`](https://github.com/hannobraun/Fornjot/blob/39fadcd9fe5effd839bf99f64d88efe672b975ea/crates/fj-window/src/run.rs#L35) to control model loading and to display those status messages in response to the model events. A [`ChangeDetected`](https://github.com/hannobraun/Fornjot/blob/39fadcd9fe5effd839bf99f64d88efe672b975ea/crates/fj-host/src/evaluator.rs#L74) event is emitted by `Evaluator` every time a model evaluation is triggered, which [results in that status message](https://github.com/hannobraun/Fornjot/blob/39fadcd9fe5effd839bf99f64d88efe672b975ea/crates/fj-window/src/event_loop_handler.rs#L60-L64).

I've been thinking through how things could be restructured to achieve this. One possibility is for `Watcher` to provide its own channel where it sends `ChangeDetected` messages, and those events could be handled by both `Evaluator` and the main event loop in `fj-window`. Instead of triggering the first evaluation by sending an initial event, `Host` could take care of triggering `Evaluator` for the first time, or there could be separate `StartWatching`/`ChangeDetected` events. Probably, the events produced by `Watcher`/`Evaluator` would be filtered through `Host` before they are handled in `fj-window`, so `fj-window` just gets a single channel to handle events from.

As I'm thinking through this, I wonder if it makes sense to do away with all the threads and just make everything async.

zthompson47:
This PR fixes #1431 and fixes #1327 by spawning a host background thread that lives outside of the the event loop.  A handle is used inside the event loop to send new models for processing.  Status messages are sent back to the event loop via `winit::event_loop::EventLoopProxy`.


crates/fj-host/src/evaluator.rs
```diff
@@ -1,9 +1,16 @@
...
+//use crossbeam_channel::{Receiver, SendError, Sender};
+use fj_interop::processed_shape::ProcessedShape;
 
-use crate::{Error, Evaluation, Model};
+use crate::Error; // Model};
 
+// -----------------
+// This file becomes unnecessary because all the processing happens directly
+// in `host.rs`.
+// -----------------
```
hannobraun, line 11: This file should be removed then.  `ModelEvent` can be moved to `host.rs`.

crates/fj-host/src/host.rs
```diff
@@ -1,31 +1,119 @@
...
-        Ok(Self {
-            evaluator,
-            _watcher: watcher,
-        })
+    /// Create a new `Host` that will process models for an event loop.
+    pub fn new(
+        shape_processor: ShapeProcessor,
+        event_loop_proxy: EventLoopProxy<ModelEvent>,
+    ) -> Self {
+        let (command_tx, command_rx) = unbounded();
```
hannobraun, line 29: Tiny nitpick: `unbounded()` doesn't really tell you what's happening here, unless you're already familiar with how channels work. This would be a bit more clear, I think:
```suggestion
        let (command_tx, command_rx) = crossbeam_channel::unbounded();
```

crates/fj-host/src/host.rs
```diff
@@ -1,31 +1,119 @@
...
+        Self {
+            shape_processor,
+            event_loop_proxy,
+            command_tx,
+            command_rx,
+        }
+    }
+
+    /// Run a background thread to watch for updates and process a model.
+    pub fn spawn(mut self) -> HostHandle {
```
hannobraun, line 39: Seems a bit superfluous, to have a separate `spawn` method. Maybe it would make for a nicer API to merge `spawn` into `new` rename `HostHandle` to `Host`, and possibly have a `HostInternal` (or whatever; I'm not married to that name) for those private methods.

crates/fj-host/src/host_handle.rs
```diff
@@ -0,0 +1,82 @@
...
+        host_thread: JoinHandle<Result<(), EventLoopClosed>>,
+    ) -> Self {
+        Self {
+            command_tx,
+            host_thread: Some(host_thread),
+            model_loaded: false,
+        }
+    }
+
+    /// Send a model to the host for evluation.
```
hannobraun, line 27:
```suggestion
    /// Send a model to the host for evaluation.
```

crates/fj-host/src/host_handle.rs
```diff
@@ -0,0 +1,82 @@
...
+/// A handle for sending commands to a spawned host
+pub struct HostHandle {
+    command_tx: Sender<HostCommand>,
+    host_thread: Option<JoinHandle<Result<(), EventLoopClosed>>>,
+    model_loaded: bool,
+}
+
+impl HostHandle {
+    /// Create a `HostHandle` with a send channel and the host thread handle.
+    pub fn new(
```
hannobraun, line 16: Is this part of the public API? Maybe `pub(crate)` is more appropriate?

crates/fj-host/src/host_handle.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// Whether a model has been sent to a host for watching and evaluation
+    pub fn is_model_loaded(&self) -> bool {
+        self.model_loaded
+    }
+
+    /// Check if the host thread has exited with a panic. This method runs at
+    /// each tick of the event loop. Without an explicit check, an operation
+    /// will appear to hang forever (e.g. processing a model).  An error
+    /// will be printed to the terminal, but the gui will not notice until
+    /// a new `HostCommand` is issued on the disconnected channel.
```
hannobraun, line 44: I think this is fine, but I'd like to mention a possible alternative: We could wrap the the host thread in [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) and send an error message through the channel, if it panics.

crates/fj-host/src/model.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::result_large_err)]
```
hannobraun, line 1: I think this is either to broad or too specific. We could decide that this warning is unreasonable. If that were the case, we should allow if for the whole crate. If we think the warning is reasonable, in general, then we should be more specific about disabling it. As it stands, it's unclear why this `#[allow(...)]` is here, and what specifically is affects.

crates/fj-host/src/watcher.rs
```diff
@@ -1,8 +1,11 @@
-use std::{collections::HashSet, ffi::OsStr, path::Path, thread};
+#![allow(clippy::result_large_err)]
```
hannobraun, line 1: See previous comment about `#![allow(...)]`.

crates/fj-host/src/watcher.rs
```diff
@@ -13,12 +16,13 @@ impl Watcher {
     /// Watch the provided model for changes
     pub fn watch_model(
         watch_path: impl AsRef<Path>,
-        evaluator: &Evaluator,
+        //evaluator: &Evaluator,
```
hannobraun, line 19: The commented code in this module can be removed, I think.

crates/fj-window/src/event_loop_handler.rs
```diff
@@ -99,8 +59,42 @@ impl EventLoopHandler {
...
+        let input_event = input_event(
+            &event,
+            &self.window,
+            &self.held_mouse_button,
+            &mut self.viewer.cursor,
+            self.invert_zoom,
+        );
+        if let Some(input_event) = input_event {
+            self.viewer.handle_input_event(input_event);
+        }
```
hannobraun, line 71: Why did this move up here? Not saying it's a problem, just wondering why.

crates/fj-window/src/event_loop_handler.rs
```diff
@@ -183,35 +177,23 @@ impl EventLoopHandler {
 
                     let gui_state = GuiState {
                         status: &self.status,
-                        model_available: self.host.is_some(),
+                        model_available: self.host_handle.is_model_loaded(), //self.host.is_some(),
```
hannobraun, line 180:
```suggestion
                        model_available: self.host_handle.is_model_loaded(),
```

crates/fj-window/src/run.rs
```diff
@@ -3,17 +3,19 @@
 //! Provides the functionality to create a window and perform basic viewing
 //! with programmed models.
 
+#![allow(clippy::result_large_err)]
```
hannobraun, line 6: See previous comment about `#![allow(...)]`.

crates/fj-host/src/host.rs
```diff
@@ -1,31 +1,119 @@
...
+            .spawn(move || -> Result<(), EventLoopClosed> {
+                let mut model: Option<Model> = None;
+                let mut _watcher: Option<Watcher> = None;
+
+                while let Ok(command) = self.command_rx.recv() {
+                    match command {
+                        HostCommand::LoadModel(new_model) => {
+                            // There's a race condition if a prior watcher sends
+                            // `TriggerEvaluation` right before this new watcher
+                            // is created.
```
hannobraun, line 53: This probably can't happen right now, as `LoadModel` is only sent, if there isn't already an existing model. Does this sound reasonable?

hannobraun (approved):
Thank you, @zthompson47, this is great!

I tested this, and it works great. I also reviewed the code and it looks good. I left comments where things jumped out to me, but none of that is a blocker.

You said that you wanted to do more cleanup and write tests before merging. While that would be very welcome, I'd be more than happy to merge this as-is. It is a clear improvement, after all, fixing two issues!

Additional cleanup and testing can be submitted in follow-up pull requests. This would also reduce the likelihood of other pull requests causing merge conflicts here in the meantime. What do you think?

One caveat: I haven't looked at #1477 yet (will do so soon), so maybe I'll end up liking that better. However, based on our conversation in #1431, I believe that the "long-lived thread" approach from this pull request is likely to be the better one.

zthompson47:
Thanks for review and comments - very helpful!

| Additional cleanup and testing can be submitted in follow-up pull requests
That sounds good to me.  I think your comments addressed most of the cleanup I was thinking of, and the tests might end up being a bit of a project.

I'll resolve the above conversations and we can merge this in.

crates/fj-window/src/event_loop_handler.rs
zthompson47: I moved it to avoid a borrow-checker issue where `event` is partially moved when a processed shape is consumed:
```
ModelEvent::ProcessedShape(shape) => {
    self.viewer.handle_shape_update(shape);
    self.status.update_status("Model processed.");
}
```
I couldn't get the compiler suggestions to work so I just moved the borrow in `input_event(&event, ...)` above the partial move.

hannobraun: Got it, makes sense!

crates/fj-host/src/host.rs
zthompson47: I think you're right that it couldn't happen with the code as-is, so the comment was for the future if there's e.g. a "Load new model" button in the gui or something similar.  Even then it would be highly unlikely and not bad even if it did happen.  It would be:
1. Old watcher sends TriggerEvaluation
2. New watcher created and new model evaluated
3. Old TriggerEvaluation received and new model evaluated again

Do you think that's the case, and if so, is it worth leaving a comment about?

hannobraun: That makes sense!

I think it's worth expanding the comment. Can't really be too explicit about these kinds of things. Only a question of time until someone sees this and spends an hour trying to understand it :smile:

crates/fj-host/src/host_handle.rs
zthompson47: That's a good idea and would probably be a better API as it would remove the need to explicitly poll for panics. 

I did it this way to keep the existing behavior intact when there is a panic in `ShapeProcessor`.  It crashes `fj-app` completely when the shape processing is in the main thread. My thinking was that a panic is unrecoverable, but I can see how that's not always the case .  In this case the user could always edit the model and try again, if we built that kind of feature in.

Note that if the host thread crashes, the app would be in an unusable state as things are currently.  An error message could explain the situation, but might be confusing for the user. Maybe we should leave things as-is for now, and revisit error handling as a separate issue?

crates/fj-host/src/host.rs
zthompson47: Something like that crossed my mind but for some reason I was stuck to the idea that the `host` variable needed to live outside the event loop.  I implemented your suggestion and it seems a lot cleaner.  It was a fairly big commit so it will need some more review before merging.

crates/fj-host/src/host_handle.rs
zthompson47: Thanks for pointing that - I haven't done much with API encapsulation, and although I've seen `pub(crate)' in other code, I haven't thought much about what it's for.  Now I know.

I made some bigger changes to the host API from the other conversation, and I think I got the API boundaries correct now, but not completely sure.

crates/fj-host/src/model.rs
zthompson47: I'm not sure what to do about this one..  The warning ultimately stems from `fj_kernel::validate::ValidationError` being large (via `fj_host::model::Error::ShapeProcessor` that I added).

There are two identical comments about it in `fj-kernel/src/lib.rs` and `fj-operations/src/lib.rs`:
```
// I've made a simple change that put `ValidationError` over the threshold for
// this warning. I couldn't come up with an easy fix, and figured that silencing
// the warning is the most practical solution for now, as the validation
// infrastructure is in flux anyway. Maybe the problem will take care of itself.
#![allow(clippy::result_large_err)]
```

Maybe we should follow suit and copy that same comment and `#![allow(...)]` over to the `fj-host` and `fj-window` crates?

---

Another point of confusion is that I had put `#[allow(clippy::large_enum_variant)]` on `fj_host::model::Error`, but I don't get a warning when I remove it.  I don't remember now if I ever got a warning or if I was just trying something out in response to a rust-analyzer diagnostic message.  However, there is also `#[allow(clippy::large_enum_variant)]` on `fj_operations::shape_process::Error` and `fj_kernel::validate::ValidationError`, and I can remove both of those without generating a clippy warning.  Anyway, this paragraph might just be something to address as a separate issue.

crates/fj-host/src/host_handle.rs
hannobraun: To be clear, this isn't very important, in my opinion, just a small improvement. My main concern with a public constructor like this is, that a user might see that in documentation, think they're supposed to use it, then get confused about how (when actually, they should create a `Host` and get a `HostHandle` through that).

In any case, I'll take a look at your new changes later, and this conversation can be resolved.

crates/fj-host/src/model.rs
hannobraun: > There are two identical comments about it in `fj-kernel/src/lib.rs` and `fj-operations/src/lib.rs`:
> 
> ...
> 
> Maybe we should follow suit and copy that same comment and `#![allow(...)]` over to the `fj-host` and `fj-window` crates?

Sounds good to me. I don't know if that is the best solution, but it certainly is the good-enough solution to move this forward.


> Another point of confusion is that I had put `#[allow(clippy::large_enum_variant)]` on `fj_host::model::Error`, but I don't get a warning when I remove it. I don't remember now if I ever got a warning or if I was just trying something out in response to a rust-analyzer diagnostic message. However, there is also `#[allow(clippy::large_enum_variant)]` on `fj_operations::shape_process::Error` and `fj_kernel::validate::ValidationError`, and I can remove both of those without generating a clippy warning.

Maybe the recent Rust 1.66 release changed the behavior? It's also possible that `ValidationError` changed in the meantime and no longer triggers this warning. Not sure though. I've worked in that general area recently, but don't remember any specific changes to `ValidationError`.


> Anyway, this paragraph might just be something to address as a separate issue.

Sounds good. As I said in my review earlier, I'd have been happy to merge as initially submitted, and this is probably extra-ready now (still have to review the latest changes). No point in holding up a substantial improvement because there are more details to be improved (there will always be more improvements to be made).

hannobraun (approved):
Thank you, @zthompson47, those latest changes look great!

I just did a rebase to get this up-to-date with `main`, and will merge as soon as CI is green.

zthompson47:
Thanks for all the help getting that PR merged!

I never did that last edit to sort out the issue with `#![allow(clippy::result_large_err)]`.  Still getting used to the workflow on github, but sounds like I'll just wrap that up and put it in a new PR (with no issue filed).

hannobraun:
> Thanks for all the help getting that PR merged!

No problem. It was you who did all of the work!


> I never did that last edit to sort out the issue with `#![allow(clippy::result_large_err)]`. Still getting used to the workflow on github, but sounds like I'll just wrap that up and put it in a new PR (with no issue filed).

Different projects handle that kind of stuff differently. Myself, I'm pretty flexible in how to do things, as I want to encourage contributions, and I don't think it's reasonable to expect every contributor to read a thick "how we do it" handbook. That said, these are my preferences:

- Open pull requests basically are a burden for everyone involved. Mostly for their author, because as long as they stay open, they tend to accumulate merge conflicts that need to be fixed, as more changes are being made to the `main` branch. That's why I prefer to merge them as soon as I feel they are an improvement, even if they have problems.
- If I merge a pull request with problems that I feel are important, I open an issue right after merging. That way, we can be sure the problem isn't forgotten, even if it's not handled right away.
- Like open pull requests, open issues also have a cost. The more you have, the harder it becomes to stay on top of them, and the more likely they are to get out of date. At the extreme end, it becomes easier to just test the app and look at the code to figure out what can be improved, than sorting through issues.
- I think the `#![allow(clippy::result_large_err)]` thing isn't quite important enough to warrant an open issue. It would still be a welcome improvement, but it's not the kind of thing we need to track through an open issue. If you wanna do it, great! Just open another pull request. If not, someone else can improve it whenever they come across it.

PR #1477, opened 2023-01-03
Improve host model loading (on-demand background thread)

### LINKED ISSUE #1327: Shape processing is blocking event loop

### LINKED ISSUE #1431: Improve initial model loading message

zthompson47:
This PR fixes #1431 and fixes #1327 by reworking the evaluator thread.  The creation of a new thread sends a `StartWatching` message to the event loop via `winit::event_loop::EventLoopProxy`.  A separate `ChangeDetected` message is sent when a model is edited.  The `ShapeProcessor` is sent to the background thread so all processing happens outside of the event loop, preventing blocking.


hannobraun (approved):
Thank you for this pull request, @zthompson47!

I think this looks great too. I'd be happy to merge it, in principle, but I prefer the approach of #1476, for the reasons we talked about in #1431.

So unless you disagree, I suggest we close this and move ahead with #1476. What do you think?

Issue #100, opened 2022-01-28
Support arcs in sketches


hannobraun:
As of this writing, sketches only support straight edges (i.e. lines segments). The CAD kernel already supports circles though, so adding support for arcs to sketches should not be that difficult.

The way sketches are defined would have to change, of course. Right now they are defined as just a list of points.

- The starting point of each edge is implicit. That prevents redundancy in the definition and ensures the sketch always forms a valid cycle.
- Hence, each edge is just defined by its end point, what type of edge it is, and edge-type-specific data required to fully define the edge.
- Line segments are just defined by their end point.
- Arcs are defined by their center point and angle (the radius can be inferred, since center and start point are known).

`fj::Sketch` could provide a convenient API:
``` rust
let mut sketch = fj::Sketch::new();
sketch.line(p1); // line starts at `p2`, the last point defined below
sketch.arc(center, angle); // arc starts at `p1` and arcs around center; distance and direction defined by `angle`
sketch.line(p2); // line starts at the end point of the arc
```

It doesn't need to look exactly like that, of course. I'm just sketching (no pun intended).

hannobraun:
This feature should ideally be able to fully replace `fj::Circle`. I just noticed that the representation I proposed above would not allow for that though.

If we can come up with something better before or while implementing this, great. If not, good enough, this would still be an improvement. In that case we should open a follow-up issue though, for fully replacing `fj::Circle` with `fj::Sketch`.

Psykopear:
I think you should also specify whether the arc is clockwise or not

hannobraun:
> I think you should also specify whether the arc is clockwise or not

Thanks for the note! That would be defined by the sign of the `angle` argument (positive angle => counter-clockwise, negative angle => clockwise, presumably).

Psykopear:
yes that works. I'm just skimming through the project, but it's really interesting, and I like the idea of something like openSCAD but with Rust code, keep up with it :)

hannobraun:
Thank you, @Psykopear!

antonok-edm:
> This feature should ideally be able to fully replace `fj::Circle`. I just noticed that the representation I proposed above would not allow for that though.

[...]

PR #1478, opened 2023-01-04
Remove default `Angle` wrapping behavior

antonok-edm:
see: https://github.com/hannobraun/Fornjot/issues/100#issuecomment-1369584250, and the commit message here


hannobraun (approved):
Thank you, @antonok-edm! The change looks great, and your reasoning makes perfect sense. Merging!

PR #1534, opened 2023-01-24
Return message string from model panic handler to be displayed in UI …

### LINKED ISSUE #1519: Panic in model crashes the app
Currently, if model code panics, this crashes the whole app:
```
thread '<unnamed>' panicked at 'not yet implemented', models/test/src/lib.rs:7:5
not yet implemented
Aborted (core dumped)
```

This can be easily reproduced by adding a `todo!()`/`panic!()`/... to any model.

Panics can be a normal part of creating a model. They can happen accidentally, or even on purpose, when placing a `todo!()` somewhere. The app should handle that case well, and just show an error message under "Status".

`fj-host`, which is responsible for handling models, already expects that any interaction with a model can result in an error, and reports those errors to the GUI. I believe that no changes are required there. So we just need to make sure that whenever model code panics, this panic is converted into an error and returned to `fj-host`.

(It would be advantageous to instead catch panics on the host's side, to protect against malicious models. But I consider such protection a nice-to-have at this point, and out of scope of this issue. In addition, we *need* to catch those panics on the model side, as panics over an FFI boundary are undefined behavior, as far as I know. Maybe we can do better [once we switch to WebAssembly](https://github.com/hannobraun/Fornjot/issues/71).)

It looks like panics in the model [are already handled](https://github.com/hannobraun/Fornjot/blob/9021f6f436016c1ef70245fa46e1d545786514ce/crates/fj/src/abi/mod.rs#L119-L132), so all we need to do is check where that handler function is called, determine if it's practical to return an error there instead, and adapt the code to do so.

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as the scope of this problem is limited to a specific part of Fornjot, and knowledge about the rest of the system shouldn't be required to address this.

mxdamien:
…status

closes #1519


crates/fj/src/abi/context.rs
```diff
@@ -23,7 +23,9 @@ impl<'a> From<&'a &dyn crate::models::Context> for Context<'a> {
             })) {
                 Ok(Some(arg)) => StringSlice::from_str(arg),
                 Ok(None) => StringSlice::from_str(""),
-                Err(payload) => crate::abi::on_panic(payload),
+                Err(payload) => StringSlice::from_str(
+                    crate::abi::on_panic(payload).to_string().as_str(),
```
hannobraun, line 27: This is unsound: `on_panic` returns an `ffi_safe::String`, which is converted into a regular `String`, which is then dropped. The `StringSlice` that references it is returned from the function, meaning the the `StringSlice` outlives the `String` it references. Also see [`StringSlice`'s documentation](https://github.com/hannobraun/Fornjot/blob/71d16ce6fb3a6e612c0e89bf46eee7278a5d1c54/crates/fj/src/abi/ffi_safe.rs#L259-L262).

---

In addition, this returns the error message just like it would return the value of the parameter. If I read the code right, this error message is then parsed as a value in the expanded model code ([[1]](https://github.com/hannobraun/Fornjot/blob/71d16ce6fb3a6e612c0e89bf46eee7278a5d1c54/crates/fj-proc/src/expand.rs#L137), [[2]](https://github.com/hannobraun/Fornjot/blob/71d16ce6fb3a6e612c0e89bf46eee7278a5d1c54/crates/fj-proc/src/expand.rs#L145)).

I think this is semi-okay in most cases, as it will cause a parse error. It might still be confusing to the user though. But it gets worse, if the parameter is a `String`. Imagine you're making a part that you want to engrave a custom message into, and you are engraving the error message instead. We're entering JavaScript territory here :smile: 

---

I think we can solve both issues by doing the following:

1. Return a `Result` (probably [`ffi_safe::Result`](https://github.com/hannobraun/Fornjot/blob/71d16ce6fb3a6e612c0e89bf46eee7278a5d1c54/crates/fj/src/abi/ffi_safe.rs#L215-L221) is required) from this method. The `Err` variant can just be the `ffi_safe::String` returned from `on_panic`, for a start.
2. Adapt all methods that call this, as necessary, returning `Result` where needed. Specifically [`Context::get_argument`](https://github.com/hannobraun/Fornjot/blob/71d16ce6fb3a6e612c0e89bf46eee7278a5d1c54/crates/fj/src/abi/context.rs#L39-L56) probably needs a custom error type, to distinguish between "argument not found" and "panicked" cases.
3. The generated model code already handles parse errors (here are the links to that expand code again: [[1]](https://github.com/hannobraun/Fornjot/blob/71d16ce6fb3a6e612c0e89bf46eee7278a5d1c54/crates/fj-proc/src/expand.rs#L137), [[2]](https://github.com/hannobraun/Fornjot/blob/71d16ce6fb3a6e612c0e89bf46eee7278a5d1c54/crates/fj-proc/src/expand.rs#L145)). We can adapt that code to handle errors resulting from a panic. ( I found it helpful to expand the macro code, for example with `cargo expand -p spacer` for the spacer model, to understand what's happening.)

crates/fj/src/abi/model.rs
```diff
@@ -46,7 +46,11 @@ impl From<Box<dyn crate::models::Model>> for Model {
                 model.metadata()
             })) {
                 Ok(meta) => meta.into(),
-                Err(payload) => crate::abi::on_panic(payload),
+                Err(payload) => {
+                    model.metadata().name =
+                        crate::abi::on_panic(payload).into();
```
hannobraun, line 51: I think we should adapt this method to return a `Result` and return an `Err` if a panic happens. Just parking the error message in the metadata, where it might not be noticed, is way too error-prone.

hannobraun (changes requested):
Thank for you for this pull request, @mxdamien!

I think this is a good start. The main cause of panics, the user-written model code, is taken care of. However, there are some problems around the periphery. I've left some comments with my thoughts. Please let me know what you think!

Also, and this is more a note to myself, here's what a `todo!()` in model code looks like, as of this pull request:
Screenshot from 2023-01-24 13-07-23

I think this is good enough for now, but eventually we would want an error message that makes clear that the error originated from a panic in the model code, and add as much detail as possible (file name, line, etc.).

None of this needs to be a part of this pull request. I'm just making a note, so I remember to open an issue, once this is merged.

hannobraun:
Thank you, David!

mxdamien:
Hi !

I hope I've addressed your remarks. I'm not entirely sure about the usage of the different internal / ffi-boundary structs. It's my first time dealing with this kind of stuff in Rust.

David

hannobraun:
Thank you for keeping at it, David! I'll take a look later today.

crates/fj/src/abi/context.rs
```diff
@@ -47,9 +61,17 @@ impl crate::models::Context for Context<'_> {
...
+            match name.trim().is_empty() {
+                true => None,
+                false => match get_argument(user_data, name) {
+                    super::ffi_safe::Result::Ok(other) => {
+                        match other.is_empty() {
+                            true => None,
+                            false => Some(other.into_str()),
+                        }
+                    }
+                    super::ffi_safe::Result::Err(_) => None,
```
hannobraun, line 73: It's not ideal to swallow the error here, as that could cause bugs to be hidden. Instead of getting an error message, making it clear that a bug has occurred, it could lead to a specified argument just not being recognized, which would be very confusing.

I think `Context::get_argument` should also return `Result` instead of an `Option`.

hannobraun (approved):
Thank you, @mxdamien! This is looking great!

I've left one comment about a thing I don't like, but I decided to merge the pull request anyway. Panics in the model code are properly handled now, and I believe that the error being swallowed there is a corner case.

I'm going to open an issue to track that. Feel free to follow up with another pull request to address it, if you want.

PR #1551, opened 2023-01-31
Fixed Status message window

### LINKED ISSUE #1508: Status messages can take up too much space
When there is a lot of information to display, the status messages will grow to take up the whole window:

Screenshot from 2023-01-13 11-59-29

This isn't great. Especially, as in this screenshot, when I already have a different source to see these errors (my IDE). I think it would probably best, if the status message field had either a fixed size, resizable by the user but otherwise constant. Alternatively, it could grow up to a maximum size that is a fraction of the screen.

Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this kind of change to the GUI is probably limited to [`Gui::update`](https://github.com/hannobraun/Fornjot/blob/55bb16e8198073c14c6eb378a906c3bd8bec0497/crates/fj-viewer/src/gui.rs#L84-L285), and won't require too much Fornjot-specific knowledge.

tmayoff:
Recently came across this project and wanted to contribute, it's not much but this should be improve the status message and close #1508. Let me know if there's anything you'd like me to change. 

image
image

### DIFF
--- a/crates/fj-viewer/src/gui.rs
+++ b/crates/fj-viewer/src/gui.rs
@@ -237,18 +237,20 @@ impl Gui {
             ui.add_space(16.0);
         });
 
-        egui::Window::new("Status").min_width(400.0).min_height(200.0).collapsible(true).resizable(true).show(&self.context, |ui| {
-            egui::ScrollArea::both().show(ui, |ui| {
-                ui.add(egui::Label::new(
-                        egui::RichText::new(format!(
-                                "Status:{}",
-                        state.status.status()
+        egui::Window::new("Status")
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @tmayoff, looks awesome!

> it's not much

Even small contributions are extremely valuable! I simply don't have the time to do everything, and there's a lot I don't know (for example, I'm not super-familiar with egui, and probably wouldn't have thought to use `egui::Window` here).

Lots of things that are good in Fornjot today exist because of small contributions that were later build upon by others!

PR #1554, opened 2023-02-02
Version Mismatches in GUI

### LINKED ISSUE #1503: Show version mismatch warning in GUI
When loading a model, Fornjot checks the version of the `fj` crate that the model was compiled against, to compare it against the version of the Fornjot app. If there is a mismatch in the "package version" (for example, the version of the Fornjot app is 0.31.0, but the model uses `fj` version 0.30.0), this leads to an error that is displayed in the GUI.

On top of that, the app also checks the "full version", which includes the Git commit hash. This is useful, if either the app or the model use a development version instead of an official release. Since this is not necessarily a problem, and also not a case that most end users will have to deal with, this only results in a warning.

The problem is, this warning is only [printed to stderr](https://github.com/hannobraun/Fornjot/blob/cdaceb97a65e69fd89b94ef3ff14c98c22a0c8c4/crates/fj-host/src/model.rs#L135-L138), which means the user might never see it. If that mismatch does cause a problem, this could be very confusing. Therefore, the warning should be shown in the GUI, maybe in addition to being printed to stderr.

The warning is printed in `Model::evaluate`, which is called by [`HostThread::process_model`](https://github.com/hannobraun/Fornjot/blob/cdaceb97a65e69fd89b94ef3ff14c98c22a0c8c4/crates/fj-host/src/host_thread.rs#L90-L112). `HostThread` already send events to the GUI via [`ModelEvent`](https://github.com/hannobraun/Fornjot/blob/cdaceb97a65e69fd89b94ef3ff14c98c22a0c8c4/crates/fj-host/src/host_thread.rs#L124-L141). To show this warning in the GUI, at least the following would have to be done:

1. Add a mechanism to tell `HostThread::process_model` about the warning. Maybe by adding a `warning: Option<String>` to [`Evaluation`](https://github.com/hannobraun/Fornjot/blob/cdaceb97a65e69fd89b94ef3ff14c98c22a0c8c4/crates/fj-host/src/model.rs#L173-L183), which is the return value of `Model::evaluate`.
2. Add a variant to `ModelEvent` and send that in `HostThread`.
3. Update the [handling of `ModelEvent`](https://github.com/hannobraun/Fornjot/blob/cdaceb97a65e69fd89b94ef3ff14c98c22a0c8c4/crates/fj-window/src/event_loop_handler.rs#L75-L97) accordingly.

I believe these steps should be relatively complete, which means the work that needs to be done is relatively limited in scope. Labeling https://github.com/hannobraun/Fornjot/labels/good%20first%20issue to reflect that.

tmayoff:
closes #1503

### DIFF
--- a/crates/fj-host/src/host_thread.rs
+++ b/crates/fj-host/src/host_thread.rs
@@ -100,6 +100,11 @@ impl HostThread {
 
         self.send_event(ModelEvent::Evaluated)?;
 
+        // TODO Following the steps exactly causes this awkwardness not all warnings are version mismatches, maybe create a Warning enum
+        if let Some(warn) = evaluation.warning {
+            self.send_event(ModelEvent::VersionMismatch(warn))?;
+        }
+
         match self.shape_processor.process(&evaluation.shape) {
             Ok(shape) => self.send_event(ModelEvent::ProcessedShape(shape))?,
 
@@ -135,6 +140,9 @@ pub enum ModelEvent {
 
     /// The model has been processed
     ProcessedShape(ProcessedShape),
+
+    /// The model was compiled against a different fj version
+    VersionMismatch(String),
 
     /// An error
     Error(Error),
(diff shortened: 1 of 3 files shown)


hannobraun (approved):
Thank you, @tmayoff! Looks and works great!

Screenshot from 2023-02-02 11-41-22

PR #1584, opened 2023-02-13
Don't require an aspect ratio when creating a new NavigationCubeRenderer

erenoku:
When the `NavigationCubeRenderer` is created for the first time just use an identity matrix as the `mvp_matrix_buffer`.


hannobraun (approved):
Thank you, @erenoku! Nice simplification!

PR #1618, opened 2023-02-24
Fix wrong link

Jzow:
The previous link should have been changed, resulting in 404 links in `README.md` under the **automation** directory


hannobraun (approved):
Thank you, @Jzow, good catch!

PR #1686, opened 2023-03-16
Add Cycle validation

A-Walrus:
Add basic checks to ensure that `Cycle`s contain at least 2 half-edges, and that all half-edges are connected to each-other.

### DIFF
--- a/crates/fj-kernel/src/objects/full/edge.rs
+++ b/crates/fj-kernel/src/objects/full/edge.rs
@@ -86,6 +86,16 @@ impl HalfEdge {
         self.curve.point_from_path_coords(start)
     }
 
+    /// Compute the surface position where the half-edge ends
+    pub fn end_position(&self) -> Point<2> {
+        // Computing the surface position from the curve position is fine.
+        // `HalfEdge` "owns" its end position. There is no competing code that
+        // could compute the surface position from slightly different data.
+
+        let [_, end] = self.boundary;
+        self.curve.point_from_path_coords(end)
+    }
+
     /// Access the vertex from where this half-edge starts
     pub fn start_vertex(&self) -> &Handle<Vertex> {
         &self.start_vertex
--- a/crates/fj-kernel/src/objects/full/edge.rs
+++ b/crates/fj-kernel/src/objects/full/edge.rs
@@ -86,16 +86,6 @@ impl HalfEdge {
         self.curve.point_from_path_coords(start)
     }
 
-    /// Compute the surface position where the half-edge ends
-    pub fn end_position(&self) -> Point<2> {
-        // Computing the surface position from the curve position is fine.
-        // `HalfEdge` "owns" its end position. There is no competing code that
-        // could compute the surface position from slightly different data.
-
-        let [_, end] = self.boundary;
-        self.curve.point_from_path_coords(end)
-    }
-
     /// Access the vertex from where this half-edge starts
     pub fn start_vertex(&self) -> &Handle<Vertex> {
         &self.start_vertex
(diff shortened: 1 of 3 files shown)


crates/fj-kernel/src/objects/full/edge.rs
```diff
@@ -86,6 +86,16 @@ impl HalfEdge {
         self.curve.point_from_path_coords(start)
     }
 
+    /// Compute the surface position where the half-edge ends
+    pub fn end_position(&self) -> Point<2> {
+        // Computing the surface position from the curve position is fine.
+        // `HalfEdge` "owns" its end position. There is no competing code that
+        // could compute the surface position from slightly different data.
```
hannobraun, line 93: This comment is completely false, and dangerously so. It used to be true for the *start* position (from where it was adapted), but by adding this method, you add the competing code that the comments in both methods claim doesn't exist.

This is a problem, because sooner or later someone is going to call this method to compare it to the start position of the next half-edge in the cycle, or use it in some scenario where it's assumed that both positions are going to be equal.

I think this method shouldn't exist. It's still possible to make the same mistake doing exactly what the method does internally, but at the very least we shouldn't make that easy. (And there should be better documentation around this, but that's a separate issue.)

However, what the new validation check does is perfectly fine, as it uses an epsilon value (`config.identical_max_distance`) for the comparison. I think it would be best to inline the body of this method into the validation check.

crates/fj-kernel/src/validate/cycle.rs
```diff
@@ -1,12 +1,132 @@
...
+    fn check_half_edges_disconnected(
+        cycle: &Cycle,
+        config: &ValidationConfig,
+        errors: &mut Vec<ValidationError>,
+    ) {
+        // If there are no half edges
+        if cycle.half_edges().next().is_none() {
+            errors.push(Self::NotEnoughHalfEdges.into());
+            return;
+        }
```
hannobraun, line 58: This is fine as-is, but I think it would be even better, if this check had its own method. Then there would be an easy 1:1 relationship between error conditions and methods. I think this would be slightly clearer.

crates/fj-kernel/src/validate/cycle.rs
```diff
@@ -1,12 +1,132 @@
...
+        for (first, second) in cycle
+            .half_edges()
+            // Chain the first half_edge so that we make sure that the last connects to the first.
+            // This unwrap will never fail because we checked before that there are enough half_edges.
+            .chain(std::iter::once(cycle.half_edges().next().unwrap()))
+            .tuple_windows()
```
hannobraun, line 64: Also fine as-is, but there's an easier (and less panicky) way to do the same thing:
```suggestion
            .circular_tuple_windows()
```

crates/fj-kernel/src/validate/cycle.rs
```diff
@@ -1,12 +1,132 @@
...
+
+        assert!(matches!(
+            invalid.validate_and_return_first_error(),
+            Err(ValidationError::Cycle(
+                CycleValidationError::HalfEdgesDisconnected { .. }
+            ))
+        ));
+
+        Ok(())
     }
```
hannobraun, line 131: Not strictly necessary, but it would be nicer to also have a test that checks for the second validation error (`NotEnoughHalfEdges`).

hannobraun (changes requested):
Thank you for this pull request, @A-Walrus! We used to have a validation check for that, but it seems like it got lost during my recent clean-up, and I missed that. Thank you for adding it back!

I've left some comments. Only one of them is critical (the one about `HalfEdge::end_position`). The other ones would be nice-to-have, but I'd be perfectly happy to merge this PR without them being addressed.

hannobraun (changes requested):
Thank you for the changes, @A-Walrus!

Looks like `HalfEdge::end_position` still exists in the final version. If you can remove it, then this is ready to merge.

Issue #42, opened 2022-01-19
Implement union operation


hannobraun:
It should be possible to create the union of two shapes, using constructive solid geometry (CSG). This is currently not supported, except for the simplest case, disjoint bodies, which is supported in the form of `fj::Group`.

Implementing this requires more solid infrastructure in the host application for storing and querying shapes. Also see #43 and #44.

hannobraun:
Blocked on #97.

hannobraun:
Now that `fj::Union` has been renamed to `fj::Group`, which explicitly is only intended for disjoint bodies, this can no longer be classified as a bug. I've updated the issue accordingly.

hannobraun:
This is no longer blocked on #97!

hannobraun:
I'm back to working on this issue directly. I'm not aware of any blockers, but of course some new ones may show up, as I'm getting into this.

hannobraun:
My attention is currently diverted to #568. While not strictly a blocker, addressing that issue will make the implementation of the union operation easier, as help as circumvent #567, which would otherwise affect the implementation.

hannobraun:
I'm still actively working on this. I made some slow progress over the last few weeks, implementing building blocks that the union algorithm is going to need. Over the last few days I've been stumbling a bit, as it became increasingly unclear what the next step is, and a few things I've tried didn't work out.

So I sat down, did some hard thinking (not easy in this heat :smile:), and came up with a plan for what's missing to take the next step with implementing the union algorithm.

So here's what I'm likely going to work on for the next weeks/months(/years? :scream:):

- [ ] Implement ray/face intersection test
  - Look into using [`robust-predicates`](https://crates.io/crates/robust-predicates), specifically `orient3d`.
  - Reuse the ray type from the triangulation/polygon code.
  - Return where the ray hits the face: vertex, edge, or face itself.
- [ ] Extract `Shell` from `Solid`
  To make sense of the ray/face intersection result, I'm going to need APIs for determining the neighbors of a face, or the faces connected by an edge. I think `Shell` would be the right place for this kind of thing.
- [ ] Implement `Solid`/`GlobalVertex` containment test
  This is going to be required for implementing `Solid`/`Edge` containment testing.
- [ ] Implement `Solid`/`Edge` containment test
  This is going to be required for implementing `Solid`/`Face` containment testing.
- [ ] Implement `Solid`/`Face` containment test
  This is going to be required required to determine which faces can be thrown away by the union algorithm.
[...]

PR #1695, opened 2023-03-19
Shell and Solid validation

### LINKED ISSUE #1594: Validate that coincident `HalfEdge`s refer to same `GlobalEdge`
If two `HalfEdge`s are coincident (which happens where two neighboring faces touch), they need to refer to the same `GlobalEdge`. If they don't, code working with the object graph (like the approximation code) won't know that they are supposed to be the same edge. In the case of the approximation code, this might lead to an invalid triangle mesh being generated. This has been the cause of bugs in the past (https://github.com/hannobraun/Fornjot/issues/494, https://github.com/hannobraun/Fornjot/issues/1162).

To make sure that all instances of this bug have been fixed, and that it stays fixed, we need a validation check for this.

A-Walrus:
Attempt to validate `Shell`'s by making sure that each half-edge in a shell refers to a global-edge that is referred to by another half-edge in that same shell. My logic says that each edge in a shell should be a part of exactly two faces but I may be wrong.

This validation check currently fails with `cargo run --package export- validator`, but I'm not sure if this is a bug in my validation checks, or whether I exposed a bug in existing code.

For example looking at the cuboid example, many global edges are referred to by only one half_edge, which seems wrong to me...

edit by @hannobraun:
Close #1594

### DIFF
--- a/crates/fj-kernel/src/algorithms/sweep/mod.rs
+++ b/crates/fj-kernel/src/algorithms/sweep/mod.rs
@@ -11,7 +11,7 @@ use std::collections::BTreeMap;
 use fj_math::Vector;
 
 use crate::{
-    objects::{Objects, Vertex},
+    objects::{GlobalEdge, Objects, Vertex},
     services::Service,
     storage::{Handle, ObjectId},
 };
@@ -47,4 +47,6 @@ pub trait Sweep: Sized {
 pub struct SweepCache {
     /// Cache for global vertices
     pub global_vertex: BTreeMap<ObjectId, Handle<Vertex>>,
+    /// Cache for global edges
+    pub global_edge: BTreeMap<ObjectId, Handle<GlobalEdge>>,
 }
--- a/crates/fj-kernel/src/algorithms/sweep/vertex.rs
+++ b/crates/fj-kernel/src/algorithms/sweep/vertex.rs
@@ -26,7 +26,11 @@ impl Sweep for Handle<Vertex> {
             .clone();
 
         let vertices = [a, b];
-        let global_edge = GlobalEdge::new().insert(objects);
+        let global_edge = cache
+            .global_edge
+            .entry(self.id())
+            .or_insert_with(|| GlobalEdge::new().insert(objects))
+            .clone();
 
         // The vertices of the returned `GlobalEdge` are in normalized order,
         // which means the order can't be relied upon by the caller. Return the
--- a/crates/fj-kernel/src/algorithms/sweep/edge.rs
+++ b/crates/fj-kernel/src/algorithms/sweep/edge.rs
@@ -37,8 +37,8 @@ impl Sweep for (&HalfEdge, &Handle<Vertex>, &Surface, Option<Color>) {
                 [
                     Some(edge.global_form().clone()),
                     Some(edge_up),
-                    Some(edge_down),
                     None,
+                    Some(edge_down),
                 ],
             )
         };
--- a/crates/fj-kernel/src/builder/cycle.rs
+++ b/crates/fj-kernel/src/builder/cycle.rs
@@ -44,9 +44,10 @@ impl CycleBuilder {
         let half_edges = edges
             .into_iter()
             .circular_tuple_windows()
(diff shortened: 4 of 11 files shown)


hannobraun:
Thank you, @A-Walrus! I'll take a closer look on Monday. For now, some comments.


> My logic says that each edge in a shell should be a part of exactly two faces but I may be wrong.

No, I think that should be correct! At least it matches what I've been thinking.


> For example looking at the cuboid example, many global edges are referred to by only one half_edge, which seems wrong to me...

Yes, that is definitely wrong. Unless there's a problem with the new validation code, this is likely a bug in the sweep code. I'm pretty sure that the sweep code creates the correct vertices (I've had problems with exactly that recently, and I'm pretty sure that's dealt with), but it might be creating duplicate `GlobalEdge`s. As long as those refer to the correct vertices, it's plausible that all the code we currently have still works correctly.

So yeah, totally plausible you exposed a bug!

hannobraun:
I've taken a look at the sweep code, and I think I've found a bug:
https://github.com/hannobraun/Fornjot/blob/ccebda5e0adb248deb52aa51a1c63901540da27e/crates/fj-kernel/src/builder/cycle.rs#L48-L49

This is missing a call to `HalfEdgeBuilder::with_global_form`. That this call is missing definitely explains the validation failure.

Problem is though, if I add the call, validation still fails. I don't know what that means. Either there are more cases like this (I couldn't find any), or something weird is going on.

A-Walrus:
There seems to have been a couple problems:
- What you found about `with_global_form`
- Incorrect order in `global_edges`
- Needing to cache edges like we do vertices

crates/fj-kernel/src/algorithms/sweep/mod.rs
```diff
@@ -47,4 +47,6 @@ pub trait Sweep: Sized {
 pub struct SweepCache {
     /// Cache for global vertices
     pub global_vertex: BTreeMap<ObjectId, Handle<Vertex>>,
+    /// Cache for global edges
+    pub global_edge: BTreeMap<ObjectId, Handle<GlobalEdge>>,
```
A-Walrus, line 51: Maybe we should have a more general caching solution. This seems to work for now but IDK if it's ideal.

hannobraun: I don't know, this seems fine to me. If you come up with something better, feel free to suggest it of course, but I think we can keep it as-is.

crates/fj-kernel/src/algorithms/sweep/vertex.rs
```diff
@@ -26,7 +26,11 @@ impl Sweep for Handle<Vertex> {
             .clone();
 
         let vertices = [a, b];
-        let global_edge = GlobalEdge::new().insert(objects);
+        let global_edge = cache
+            .global_edge
+            .entry(self.id())
+            .or_insert_with(|| GlobalEdge::new().insert(objects))
+            .clone();
```
hannobraun, line 33: I swear, I was looking for exactly this code yesterday. I found the vertex caching code instead, and my brain was like "oh, so we do have the required caching". I'm pretty stupid sometimes :smile: 

In any case, good find!

hannobraun (dismissed):
Thank you, @A-Walrus, this is great!

> There seems to have been a couple problems:
> 
> * What you found about `with_global_form`
> * Incorrect order in `global_edges`
> * Needing to cache edges like we do vertices

This is some nice debugging. Well done!

I left some comments, but there's nothing critical. There's one suggestion that I will apply myself after submitting this review. This is good to merge!

Some additional comments:

- In general, I'd prefer if the final order of commits were such, that every single commit works (that means moving the fixes before the validations that require them). This doesn't matter most of the time, but if you're using `git bisect`, then a commit that has build errors or test failures can be inconvenient.
- As you mention, the test cases only cover failure. This would be hard to fix right now, as we don't have good tools to build `Shell`s and `Solid`s. But since I'm going to [focus on APIs to build shapes](https://github.com/hannobraun/Fornjot/issues/42#issuecomment-1473709945) for the foreseeable future, those test cases are probably a good testing ground for those new APIs.

PR #1696, opened 2023-03-19
Update docs for `Solid`

A-Walrus:
Based on discussion at #1694 updated the docs to make it a bit clearer.

### DIFF
--- a/crates/fj-kernel/src/objects/full/solid.rs
+++ b/crates/fj-kernel/src/objects/full/solid.rs
@@ -6,7 +6,7 @@ use crate::{
 };
 
 /// A 3-dimensional shape, built from [`Shell`]s. Many Solids will contains only
-/// one shell, but if the Solid contains cavities they will be reprented by a
+/// one shell, but if the Solid contains cavities they will be represented by a
 /// shell each, as well as a shell for the outside.
 ///
 /// # Implementation Note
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you, @A-Walrus! I'll take a look on Monday.

crates/fj-kernel/src/objects/full/solid.rs
```diff
@@ -5,7 +5,9 @@ use crate::{
     storage::Handle,
 };
 
-/// A 3-dimensional shape
+/// A 3-dimensional shape, built from [`Shell`]s. Many Solids will contains only
+/// one shell, but if the Solid contains cavities they will be reprented by a
+/// shell each, as well as a shell for the outside.
```
hannobraun, line 10: A typo:
```suggestion
/// A 3-dimensional shape, built from [`Shell`]s. Many Solids will contains only
/// one shell, but if the Solid contains cavities they will be represented by a
/// shell each, as well as a shell for the outside.
```

Additional comments:
- I think it's better to have a short 1-line header for a doc comment, before going into details. I believe (although I haven't checked in a while) that leads to nicer generated documentation, than starting with a multi-line paragraph.
- This leaves out the possibility that the `Solid` could be made up of multiple bodies, i.e. have multiple exterior shells.

hannobraun (changes requested):
Thank you for the pull request, @A-Walrus!

I've left a comment, but the only real problem is the typo, for which I've left a suggestion. I think I can probably incorporate my suggestion myself, and if so, I will merge this immediately.

If you agree with my comments, feel free to submit a follow-up pull request, if you want!

PR #1698, opened 2023-03-19
Allow export-validator to recieve model name

A-Walrus:
If you pass export_validator a specific model name as an argument, it will only check that specific model (I found this useful for debugging)
If you pass no ARGS it will check all models like before


hannobraun (approved):
Thank you, @A-Walrus! Looks great!

PR #1710, opened 2023-03-21
Fix rust-analyzer in nix flake

A-Walrus:
Added "rust-src" component which is necessary for rust-analyzer in nix flake.

PS: Sorry for all the nix PRs, I think this is the last one...


hannobraun (approved):
Thank you, @A-Walrus!


> Sorry for all the nix PRs, I think this is the last one...

No problem, always happy to receive fixes!

Issue #1571, opened 2023-02-08
`fj-host` depends on winit


hannobraun:
The [`fj-window`](https://github.com/hannobraun/Fornjot/tree/main/crates/fj-window) crate was created to separate the choice of window library (for which we use [winit](https://crates.io/crates/winit)) from the graphics code (which lives in [`fj-viewer`](https://github.com/hannobraun/Fornjot/tree/main/crates/fj-viewer)). The goal of this separation is to allow anyone who wants to create their own app based on Fornjot to make their own choice of window library, while still being able to reuse `fj-viewer`.

During some recent work, a dependency on winit was added to [`fj-host`](https://github.com/hannobraun/Fornjot/tree/main/crates/fj-host), which runs counter to that goal. `fj-host` should not depend on winit.

The reason it does, is because it uses [`EventLoopProxy`](https://docs.rs/winit/0.28.1/winit/event_loop/struct.EventLoopProxy.html) to inject events into `fj-window`'s event loop (see [`host.rs`](https://github.com/hannobraun/Fornjot/blob/main/crates/fj-host/src/host.rs) and [`host_thread.rs`](https://github.com/hannobraun/Fornjot/blob/main/crates/fj-host/src/host_thread.rs)). I think we can replace this with a more generic interface (maybe a channel from [`crossbeam-channel`](https://crates.io/crates/crossbeam-channel), which `fj-host` already uses) and add an adapter that uses `EventLoopProxy` to `fj-window`. That would then allow anyone to use `fj-host` without having to depend on winit.

Labeling as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, since this is a fairly localized change that requires some understanding of the interaction between `fj-window` and `fj-host`, but no knowledge of the larger context of Fornjot.

PR #1712, opened 2023-03-21
Implements #1571

IamTheCarl:
Resolves https://github.com/hannobraun/Fornjot/issues/1571

I love how you've structured this project and am excited to see what it brings to the maker industry.
Gave me a few ideas for an EDA application built on the same principals but I'd like to learn a bit more about your way of doing things before diving into that project.


hannobraun (approved):
Thank you, @IamTheCarl, this looks great!


> I love how you've structured this project and am excited to see what it brings to the maker industry.

Thank you!


> Gave me a few ideas for an EDA application built on the same principals but I'd like to learn a bit more about your way of doing things before diving into that project.

The dream would be to have CAD and EDA integrated into one application/ecosystem, but I think that's a long way off. Still lots to do on the CAD side around here :smile:

Issue #1569, opened 2023-02-07
Print stack trace for panics that originate in model code


hannobraun:
Executing a `todo!()` in the code of a model results in this error message:
![Screenshot from 2023-02-07 13-00-09](https://user-images.githubusercontent.com/85732/217239505-363800e8-0f31-4df1-a750-a7581e8df2e5.png)

This is not ideal, as it lacks some critical information:

- It doesn't explain *that* a panic occurred.
- It doesn't show *where* that panic occurred.

Ideally, the error message should clearly explain that there was a panic while loading the model, originating from the model code, and print a stack trace for the panic. Even more ideally, this stack trace would leave out any irrelevant parts from outside of the model code, and only show the stack trace of the panic starting from the model function.

Panic handling for models centers around the [`on_panic`](https://github.com/hannobraun/Fornjot/blob/ed08c4c8d99d3e146a9a815406e02e276a040d77/crates/fj/src/abi/mod.rs#L121-L132) function, and I think that any necessary changes will either be completely contained within that function, or at least center around it (if changing the function's signature is necessary, that will require changes in the code using that function).

Labeling as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, as this is a fairly localized change. It will require some knowledge of Rust panic and error handling to address, but shouldn't require much knowledge about Fornjot in particular.

IamTheCarl:
Looking into the issue with the panic handler not intercepting panics caught by `catch_unwind`, it seems that the panic hook is a global variable and that each module we're loading has defined its own personal panic hook, so that's why that isn't working. We'd have to make some kind of change to the client library to fix that, or use a completely different approach to models.

IamTheCarl:
In case you're curious, [here's my branch](https://github.com/IamTheCarl/Fornjot/tree/better-model-panics) implementing what I have. Most of the changes are [here](https://github.com/IamTheCarl/Fornjot/blob/better-model-panics/crates/fj/src/abi/mod.rs#L158).

hannobraun:
Thank you for looking into this, @IamTheCarl!

> I've started to take a crack at implementing this and have run into problems.
> `catch_unwind` will not give you the information you are looking for, period.
> I'm not sure why. I think they just don't want you doing that.

That's too bad.

Maybe the issue is that the backtrace isn't captured? You have to run a Rust application with `RUST_BACKTRACE=1` to get one for panic messages, after all. If that's the case, I wonder if we can enable that unconditionally for model code somehow?


[...]

PR #1716, opened 2023-03-23
Better model panics

IamTheCarl:
Mostly ~resolves~ addresses https://github.com/hannobraun/Fornjot/issues/1569 . (edited by @hannobraun: "resolves" is a magic word that closes the issue; GitHub will readily take instructions from random words within the pull request description, but doesn't know what "mostly" means)

It currently does not print the stack trace, but it does print the fact that it's a panic and a line number where the panic occurred. Given a little more time I might be able to get the stack trace.

image

The key is to override the panic hook within the model. This cannot be done externally, so the initialization code is injected into the model function by the `#[fj::model]` macro. That new panic hook then places all panics into a global variable for the `on_panic` function to grab. I suspect this could get a bit funky when multiple threads are involved, but the worst I would expect is it giving a "no details provided" message or the wrong error message on a panic.


crates/fj-proc/src/lib.rs
```diff
@@ -91,13 +95,67 @@ pub fn model(_: TokenStream, input: TokenStream) -> TokenStream {
...
+                            },
+                        })),
+                        paren_token: Paren {
+                            span: Span::call_site(),
+                        },
+                        args: Punctuated::new(),
+                    }),
+                    syn::token::Semi::default(),
+                ),
+            );
```
hannobraun, line 149: I'm far from an expert in procedural macros and might be wrong about what's happening here, but couldn't this be done much easier using the `quote!` macro? See [`expand.rs`](https://github.com/hannobraun/Fornjot/blob/main/crates/fj-proc/src/expand.rs) for examples.

crates/fj-proc/src/lib.rs
```diff
@@ -91,13 +95,67 @@ pub fn model(_: TokenStream, input: TokenStream) -> TokenStream {
...
 
             let tokens = quote::quote! {
                 #item
                 #init
+
             };
 
+            eprintln!("TOKENS: {}", tokens);
+
```
hannobraun, line 158: It it intentional that this was left in here? Or just leftover debug output?

hannobraun (approved):
Thank you, @IamTheCarl! Looks good and works great.

I've left two comments with suggestions for potential improvements. None of those are critical, so I'm going to merge this now. But feel free to send a follow-up PR, if you agree with them.

I also made a small edit to your pull request description, to prevent #1569 from being closed when I merge this.

crates/fj-proc/src/lib.rs
IamTheCarl: Leftover debug output... I thought I got them all...

Issue #1496, opened 2023-01-10
Make status message expandable to show more detail


hannobraun:
The Fornjot app displays status messages, to show what's currently happening with the model.

![Screenshot from 2023-01-10 12-08-15](https://user-images.githubusercontent.com/85732/211535860-314f15da-3a37-407d-91a3-8799487acd30.png)

These status message provide a high-level overview, but there's more detail going on behind the scenes. Not showing this detail can be confusing. For example, when loading a model for the first time, it can take longer than expected, as the model's dependencies need to be compiled first.

At the same time, showing *all* the detail would be overwhelming. So I've come up with the following idea to address this:

- Add a UI element to the left of each status message that allows you to expand that status message, to see more detail.
- I often see triangles to represent such UI messages, with the triangle pointing to the right while the element is not expanded, then down once it is expanded. But such details don't matter for an initial implementation. If [egui](https://crates.io/crates/egui) already offers something like this, we can just use it.
- If a status message is expanded, it shouldn't scroll out of sight because too many details are displayed. It should always stay visible, to always make it clear to which message the currently displayed details belong, and to always make it possible to un-expand the details.
- I think the way to achieve that, would be to display the expanded details in a scrollable UI element that has a constant size.
- We might consider to automatically expand the latest message, and automatically un-expand it, once the next message is displayed. But this is a detail that can be left for further iteration.
- Initially, the only details displayed in this way could the the output from Cargo when compiling a model.

I haven't looked into implementing this, but off the top of my head, this is what would need to happen:

- Status messages are stored in the [`StatusReport`](https://github.com/hannobraun/Fornjot/blob/c97e5bd87663adb6e371c0b1987ab1ee09ec4763/crates/fj-viewer/src/status_report.rs) struct. This needs to be expanded, to support the new capabilities.
- `StatusReport` is accessed from [`Gui`](https://github.com/hannobraun/Fornjot/blob/c97e5bd87663adb6e371c0b1987ab1ee09ec4763/crates/fj-viewer/src/gui.rs) to be displayed. The new UI would need to be implemented there.
- `StatusReport` is stored in [`EventLoopHandler`](https://github.com/hannobraun/Fornjot/blob/c97e5bd87663adb6e371c0b1987ab1ee09ec4763/crates/fj-window/src/event_loop_handler.rs) and updated by its methods, based on the events they receive.
[...]

PR #1721, opened 2023-03-23
Better model panics

IamTheCarl:
Finishes off the last changes needed to resolve https://github.com/hannobraun/Fornjot/issues/1569
Yes it even cleans up the [syn AST stuff.](https://github.com/hannobraun/Fornjot/pull/1716#pullrequestreview-1354239897)

image

My only complaint is that the stack trace is really long but I don't think there's much we can do about that at this time.


hannobraun:
Thank you for the pull request, @IamTheCarl! I'll take a look soon (most likely later today).

A note: I generally prefer rebase over merge to fix conflicts, because all those merge commits end up making the pull request harder to review for me. A merge tends to make more sense, if I've already reviewed a pull request, but more changes were necessary (because then I can just ignore the previous commits and focus on the new commits when I review again). This is documented in [`CONTRIBUTING.md`](https://github.com/hannobraun/Fornjot/blob/main/CONTRIBUTING.md) ([[1]](https://github.com/hannobraun/Fornjot/blob/main/CONTRIBUTING.md#pull-request-lifecycle), [[2]](https://github.com/hannobraun/Fornjot/blob/main/CONTRIBUTING.md#dealing-with-changes)).


> I'm not experienced with CI so I have to ask, these checks that seem to run with every commit I make, they don't cost you extra money every time I trigger them, right?

It doesn't. CI resources are provided by GitHub for free, fortunately.

It still makes sense to keep CI usage down, to not waste resources, to not compete with other ongoing builds (number of concurrent builds per project is restricted, I believe), and not least for your own convenience (CI turnaround times are annoyingly long). You can do most of what the CI build ends up doing locally, by running `just ci` ([also documented in `CONTRIBUTING.md`](https://github.com/hannobraun/Fornjot/blob/main/CONTRIBUTING.md#running-a-full-build)).

hannobraun:
Ah, I forgot to comment on something.

> My only complaint is that the stack trace is really long but I don't think there's much we can do about that at this time.

Yeah, don't worry about that. It's better to have stack traces, because once we have them, we can improve upon them. I already have some ideas, like expandable status messages (https://github.com/hannobraun/Fornjot/issues/1496).

crates/fj-proc/src/lib.rs
```diff
@@ -148,6 +148,18 @@ pub fn model(_: TokenStream, input: TokenStream) -> TokenStream {
...
+            let sig = item.sig;
+            let statements = item.block.stmts;
+
+            let item = quote::quote! {
+                #(#attrs)* #vis #sig {
+                    fj::abi::initialize_panic_handling();
+                    #(#statements)*
+                }
+            };
+
```
hannobraun, line 162: You seem to be adding the new and improved code here, without removing the old and busted one. `initialize_panic_handling` says it can be called multiple times, but I think that's not enough reason to keep the old code around :joy:

hannobraun (changes requested):
Thank you for the pull request, @IamTheCarl, looks good!

I left one comment that I'd like to see addressed. And since I'm already requesting changes, I don't see that leftover debug message being removed. That would be nice too.

Other than that, this is good to go!

crates/fj-proc/src/lib.rs
IamTheCarl: I Had removed the old version and that println but for some reason merging brought a lot of the old stuff back. I missed those two.
I'm going to have to give this re-basing thing a try.

I'll try to fix that up today.

hannobraun (approved):

hannobraun:
Looks good to go now! Thank you, @IamTheCarl!

Issue #1722, opened 2023-03-24
Fornjot models as serialized data


ahdinosaur:
Hi,

Lovely to follow along with the great work on Fornjot, is exciting. :smiley_cat: 

I'm playing with a way of writing Fornjot models in JavaScript / Typescript using [`deno_core`](https://crates.io/crates/deno_core).

Here might be the main example model in JavaScript:

```ts
export const params = [
  {
    name: 'outer',
    type: 'number',
    default: 1.0,
    min: ({ inner }) => inner * 1.01,
  },
  {
    name: 'inner',
    type: 'number',
    default: 0.5,
    min: ({ outer }) => outer * 0.99,
  },
  {
    name: 'height',
    type: 'number',
    default: 1.0,
  },
]

export function shape(params) {
  const { outer, inner, height } = params

  const outer_edge = Sketch.from_circle(Circle.from_radius(outer))
  const inner_edge = Sketch.from_circle(Circle.from_radius(inner))

  const footprint = outer_edge.difference(inner_edge)
  const spacer = footprint.sweep([0, 0, height])

  return spacer
}
```

But I'm having trouble figuring out how to use the existing `fj-app` and `fj-host` interfaces.

As far as I understand, a shape is comprised of `fj` values, which thankfully are serializable with `serde`. A model is some (serializable) metadata (like parameter definitions) plus a shape function that is given parameter values and returns (serializable) shape data. I like this, yay code as data.

But then I see `fj-app` and `fj-host`, and I'm not sure what to do. `fj-app` expects a `Model` struct, that seems very restrictive in how it works. I would much rather implement a `Model` trait and provide the necessary interfaces (`get_parameters()`, `get_shape(parameters)`, etc) by serializing and deserializing data between Rust and v8 (JavaScript), and ignore `fj-host` altogether.

As another layer I could also imagine a "loader" trait that is given model ids (paths or urls) and returns models, and some such to track when models have changed. Then we could have a Rust loader, a WASM loader, a JavaScript loader, a Rhai loader, etc. 

But I know this topic has been discussed before, e.g. https://github.com/hannobraun/Fornjot/issues/71, https://github.com/hannobraun/Fornjot/issues/804, etc, so is very possible I'm missing something important.

Thanks :purple_heart:

IamTheCarl:
I'm actually working on this right now. Most of my idea is explained [here](https://github.com/hannobraun/Fornjot/issues/1569#issuecomment-1479870965).

[...]

Issue #1713, opened 2023-03-22
Create builder API for `Solid`/`Shell` validation tests


hannobraun:
There are validation checks for `Solid` and `Shell` whose unit tests only test the failure case of those validation checks. They don't make sure, that the validation check passes for a valid model. This makes total sense, as creating a valid `Solid`/`Shell` is a huge pain with the currently available APIs.

As part of #42, I'm starting to work on better APIs for creating shapes. It makes sense to start with an API specifically tailored to these unit tests, as those would be a nice proving ground for the API design, and the result would immediately become part of an automated test suite.

I have already started working on this.

hannobraun:
I've made some nice progress here, as the references from merged pull requests above this comment can attest. I believe the new API is very promising. There's still work to do, but I'm hopeful that it can expand to cover uses cases beyond the scope of this issue.

I'd still like to do the following, before considering this issue complete:

1. [ ] Migrate code that still uses the old builder API to the new one (which I might come to call the "operations" API; we'll see) where possible. There should be quite a bit of test code that can already be supported by the new API.
2. [ ] Merge the old builder API into the new builder/operations. The old API still covers some use cases that the old one doesn't. Merging those in would further proof out the concepts of the new API.
3. [ ] Write unit tests for the `Solid` validation checks. Those currently don't have any. Writing them will require further expansion of the new builder/operations API.

Overall, unless there are some surprises waiting, this shouldn't be much more work.

hannobraun:
I've been back from vacation this week and have gotten back into this issue. I've been working on unit tests for the `Solid` validation checks, which required improvements and fixes to the operations API.

The main thing that's still in progress is an API for joining cycles together, where their edges touch. The last approach I tried didn't work out very well (turned out too hard to use). I have another idea that I'm hopeful about though.

In the meantime, I've submitted quite a few pull requests with all kinds of improvements and additions that came out of this.

hannobraun:
I've put more work into this issue since my last comment here (as evidenced by the list of pull requests above), but then I got side-tracked and forgot to update this issue.

I believe that [my list above](https://github.com/hannobraun/fornjot/issues/1713#issuecomment-1482809021) is still accurate, which would mean the only thing left to address here, is to write unit tests for `Solid` validation. I'm not sure what, if anything, is missing from the builder API to support those tests. The API has been evolving, and should be better than ever.

[...]

PR #1735, opened 2023-03-28
Serialization based models

IamTheCarl:
This idea started in #1569 and has generally been talked about in #1722 . I do not yet suggest closing #1722 as I think there will still be a a lot more to discuss on this idea. This is just the first proof of concept.

To simplify the implementation of #71, the communications between Fornjot and its clients has been converted to structs that are serialized with serde to the postcard format. This is done for both sending arguments to the client and for the client to send error messages or the generated shape structure to Fornjot.

This vastly simplifies the ABI for Fornjot, only requiring functions in the ABI to pass or accept pointers and integers.
I plan to make more breaking changes to the ABI later but once I have the vestigial code removed from this branch, I think we should merge it.


hannobraun (commented):
Thank you for the PR, @IamTheCarl!

While I didn't find anything specifically wrong here (I left a comment for a small nitpick though), I have to admit, my eyes were glazing over quite a bit while reviewing this. Most commits contains a large amount of changed lines, a lot of that within `unsafe` functions or blocks.

Normally I'm not very strict about this kind of thing, but since we're dealing with `unsafe` (and thus, potentially unsound code) here, I'd like to be a bit more confident before merging this.

I think we have three options:

1. Just YOLO it, see how it works out. Hope that someone will find any soundness issues later, if there are any. I don't want to do this.
2. I could power through, carefully reviewing the end result line-by-line. This is an option, but I'd prefer option 3.
3. You could package your changes into smaller, more focused commits, each of which are self-contained and easy to review.

To give you some specific examples of what I'm looking for:

- Your [first commit](https://github.com/hannobraun/Fornjot/pull/1735/commits/ee7eb18e9753ce97238bb8392b570b09c7f655f1) is basically perfect: One small, focused change, explained in the commit message. I know I agree with the commit message, so the only thing I have to keep in mind while reviewing this, is to make sure the changes match that message. Pretty easy!
- Your [second commit](https://github.com/hannobraun/Fornjot/pull/1735/commits/ad9b2c475da916376a33a7d53f4e61a813132847) on the other hand, contains multiple not-closely-related changes, and I have to juggle a lot of state in my head:
  - It removes the redundant code that was replaced with a `quote!`. I happen to remember why this code is redundant and that removing it is good, so no big deal in itself, but it's one more thing to keep in mind.
  - It removes some files outright, and I'm not sure: Have those already been useless? Are they replaced by the additions in this commit? Is it even correct to remove them? Again, no big deal in itself, but it adds to the difficulty.
  - It changes the way parameters are handled. Do I like this approach? Is it flexible enough? That extra trait shouldn't be necessary, let's start writing a comment. Or is it used somewhere else? Oh, it's removed in the next commit. Guess I didn't have to think about any of that.
  - It derives more traits for various existing structs. Again, not a big deal (actually a really tiny one), but in a commit that's already hard to get a hold of, this is one more source of visual noise, and something that could have easily been done in a separate commit.
- Your [fourth commit](https://github.com/hannobraun/Fornjot/pull/1735/commits/49c4517f6f815c9a682deec2b7289692c855956e) removes quite a bit of code that could probably have been removed in a follow-up commit, making each of the resulting smaller commits easier to review.

I don't mean to pick on you or your code here. I'm just trying to explain why this is difficult and time-consuming to review. On the other hand, learning how to create small and self-contained commits is a valuable skill, and one that becomes second nature once you're used to it.

If you want to see an example of how I do it, maybe #1711 and #1714 are decent examples. I'm far from perfect, but you can see there, I try to put each change into a self-contained commit. I sometimes commit refactorings that don't make much sense by themselves, but that are easy to understand; and that make a follow-on commit smaller, and thus also easier to understand.

If you're not experienced enough with Git to know how to do this, I encourage you to look into the following commands:
- Precisely control what goes into the next commit: `git add -p`, `git stash -p`, `git stash --keep-index`
- Roll back the last few commits, without changing the actual files: `git reset --mixed HEAD~1` (replace `1` with the number of commits you want to roll back; very useful when combined with the previous item, to split a commit into smaller, more self-contained ones)
- Instead of creating a new commit, add changes to the previous one: `git commit --amend`
- Manipulate Git history (drop, merge, or re-order commits): `git rebase -i`

All of these are tools that I use pretty much daily. If you don't feel confident in your abilities yet, I recommend trying it in a separate branch, as manipulating Git history can get hairy (speaking of which, `git cherry-pick` is pretty useful when creating a new branch to re-organize what you did in another). If you truly mess up, `git reflog` can usually help you restore what you accidentally destroyed.

If you need any help with that, or don't feel up for the task, please let me know! I'm happy to help. And if you don't have the time or wherewithal to deal with this, also let me know. I can look into cleaning the pull request up myself, or just power through with the review :smile: 

> once I have the vestigial code removed from this branch, I think we should merge it.

As I said earlier, I'm happy to merge this with vestigial code still in place. Please consider, the longer this pull request stays open, the higher the chance that other pull requests will cause merge conflicts here, causing you more work to get the pull request mergeable again.

IamTheCarl:
It sounds like you would have an easier time committing this now rather than waiting for me to polish this off, so I'll remove the WIP status and let you do that whenever you feel ready.

I'll try to use a commit pattern more to your liking. These particular commits were a bit larger because larger changes were needed to get the project back into a compiling state.

hannobraun:
> It sounds like you would have an easier time committing this now rather than waiting for me to polish this off, so I'll remove the WIP status and let you do that whenever you feel ready.

Well, the code itself doesn't need more polish, but the commits could certainly use some (as described above). There are also merge conflicts with the `main` branch.

I can take care of that, but I'll have to find the time. Currently, I'm still working on #1713.


> I'll try to use a commit pattern more to your liking. These particular commits were a bit larger because larger changes were needed to get the project back into a compiling state.

Yes, sometimes large commits (and thus a difficult review) can't be avoided. All the more reason to keep things as clean as possible, to not further add to the difficulty.

IamTheCarl:
I'll check it out tonight. If I remember correctly, that backtrace was removed and replaced with some strings and integers for this very problem. The backtrace probably got brought back in the rebase.

This should be a good learning experience for me.

IamTheCarl:
Turns out I had forgotten to bring the changes merged into main into this branch. This branch was started on before the backtrace feature was merged into main.

Attempted to rebase and failed, tried a merge and that worked.
I certainly need to experiment more with this stuff. I'm wondering if it would be helpful to you to try and squash some of these commits. I'm going to try and clean up the commit history a little but if I haven't by the time you see this and you're happy with the state of the code, go ahead and merge it in anyway.

hannobraun:
> Attempted to rebase and failed, tried a merge and that worked.

Something's still not right. GitHub is showing various conflicts.


> I certainly need to experiment more with this stuff. I'm wondering if it would be helpful to you to try and squash some of these commits. I'm going to try and clean up the commit history a little but if I haven't by the time you see this and you're happy with the state of the code, go ahead and merge it in anyway.

Any history cleanup would be very welcome! I don't think I can merge this very soon. Pretty sure I won't get to it today, and I'm on vacation starting Tuesday. (Being on vacation could mean that I'll look into this pull request, if the mood strikes me, or it could mean I have no time because I'm out hiking :smile: )

So there's definitely no time pressure from my side.

PR #1739, opened 2023-03-31
OBJ file format for export

replicadse:
I tested a few things using Forgot to create some shapes.
In order to export / import these I needed the OBJ format. That's what I added right into it here and thought you might be interested in having this native inside of the application.

The export uses a crate I built a few months / years ago (wavefront_rs). OBJ format is implemented. MTL format will follow.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1072,6 +1072,7 @@ dependencies = [
  "stl",
  "thiserror",
  "threemf",
+ "wavefront_rs",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


hannobraun:
Thank you for this pull request, @replicadse! I'm in a bit of a hurry right now, but I'll be able to take a closer look on Monday.

CI is failing though. Looks like the new error variant introduced an error in `fj-app`.

replicadse:
Cool - yeah I saw that this morning. Fixing that error today.

replicadse:
@hannobraun got the pipeline fixed now - lmk if you have questions

hannobraun:
Nice! I'll review it tomorrow.

replicadse:
Thanks @hannobraun I will check it out. I also plan to make some qol updates to the wavefront library incl materials support etc.
Will make a PR accordingly and check the edges stuff. Also, I think there's probably a more optimal solution to the export than giving every triangle it's own three points (index list / vertex list?).
Thanks!

Issue #1568, opened 2023-02-07
Errors from retrieving model arguments are swallowed


hannobraun:
The Fornjot application communicates with the models it loads over an FFI boundary (with Rust code on both sides). This is necessary, because models are compiled to dynamic libraries, and Rust doesn't have a stable ABI. Panics in model code must not cross this FFI boundary, as that would be undefined behavior. Therefore, panics in model code are caught and turned into errors, which are then displayed in the Fornjot application.

There is one hole in the code that is handling this. If a panic occurs in a model's argument handling code (which is generated by `fj-proc`), it is converted into an error, but then later silently ignored:
https://github.com/hannobraun/Fornjot/blob/ed08c4c8d99d3e146a9a815406e02e276a040d77/crates/fj/src/abi/context.rs#L73

This is not ideal, as that could hide bugs in a highly confusing manner. Instead, the method this is happening in (`Context::get_argument`) should return a `Result` instead of an `Option` and pass that error on. Any downstream code should be adapted as necessary.

Labeling as https://github.com/hannobraun/Fornjot/labels/good%20first%20issue, since this is a pretty localized problem.

emcloughlin:
I think I will try to tackle this as a first issue. Wish me luck lol

hannobraun:
Sounds great! Looking forward to what you come up with!

PR #1740, opened 2023-04-02
Issue#1568 - Proper error handling for Context::get_argument()

emcloughlin:
Hi guys,

This is my first time ever submitting a pull request, so please "be gentle" lol. I'm extremely open to criticism and feedback, while my code does successfully compile I'm not at all confident in the implementation, but I figured I'd just jump in head first.

This is a fix for #1568, which should add proper error handling for problems with processing model arguments. To be fully honest, I'm really at the point where I just need to be told the code looks good, or to be told it's absolute shit and then to receive some pointers about why what I did is wrong and what I should try to do instead.

One thing to note - I accidentaly took the idea of "bite-sized" commits a little too seriously, and as a result the first two commits don't actually compile, as I hadn't even finished writing my first implementation when I committed them. Will make sure it doesn't happen again, however.

Thanks in advance! This project is something I'm really excited about and I hope this is just the first commit of many to come.


hannobraun:
Thank you for the pull request, @emcloughlin! And thank you for the honor of starting your open source career with Fornjot :smile: 

I'll take a closer look when I'm back in the office tomorrow. In the meantime, there seem to be some issues with the CI build:

1. The formatting check fails. That should be an easy fix: Run `cargo fmt`, commit.
2. One of the example models fails to compile (maybe others too; the test run aborts, if it hits an error). You might need to update them according to the changes you made.

As a general note, you can run the failing model test locally with `just test`, and you can replicate the full CI build locally (for the most part) with `just ci`. This is [documented in `CONTRIBUTING.md`](https://github.com/hannobraun/Fornjot/blob/main/CONTRIBUTING.md#running-a-full-build).

hannobraun (changes requested):
Thanks again for the pull request, @emcloughlin!

I think this looks very good. I left two comments were things could be improved, but this is fine as-is. Happy to merge this, despite these comments!

The only real problem is that the CI build is failing. See my comment from yesterday. This needs to be fixed.


> One thing to note - I accidentaly took the idea of "bite-sized" commits a little too seriously, and as a result the first two commits don't actually compile, as I hadn't even finished writing my first implementation when I committed them. Will make sure it doesn't happen again, however.

Yeah, small commits are good, but they should compile, in general. Otherwise, you can run into trouble when running `git bisect`. But it's fine, definitely not a show-stopper.

In this specific case, I'll probably just squash the commits into a single one when merging, as the overall pull request isn't that big.

Thank you again, @emcloughlin! Happy to merge this, once the CI build is green.

PR #1781, opened 2023-04-22
Turn pretty_assertions into a "dev" dependency

therealprof:


hannobraun (approved):
Thank you, @therealprof, good catch!

PR #1800, opened 2023-04-26
Fix sweep docs

### LINKED ISSUE #1799: Incorrect sweep docs?
https://github.com/hannobraun/Fornjot/blob/a4d75d87245368b4ba964f1698ad898da4f209b6/crates/fj/src/lib.rs#L49-L50

It says that it is swept along the z axis, but from what I can tell it seems that it can be in any vector (length + direction):
https://github.com/hannobraun/Fornjot/blob/a4d75d87245368b4ba964f1698ad898da4f209b6/crates/fj/src/sweep.rs#L23-L24

I'd be happy to fix the docs, unless I'm missing something and it does indeed have to be in the z-axis

A-Walrus:
Closes #1799

### DIFF
--- a/crates/fj/src/lib.rs
+++ b/crates/fj/src/lib.rs
@@ -46,7 +46,7 @@ pub enum Shape {
     /// A 2D shape
     Shape2d(Shape2d),
 
-    /// A sweep of 2-dimensional shape along the z-axis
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @A-Walrus!

Issue #1689, opened 2023-03-16
Validate that coincident vertices within `Sketch` and `Solid` are identical


hannobraun:
Shapes in the Fornjot kernel are made up of objects (faces, edges, vertices, ...). Those objects are managed in a centralized data store, and their location in that store gives them a unique identity. This means that two objects that have been created at different types are never identical, even if they are equal (i.e. contain the same data).

If two vertices within a `Sketch` or `Solid` are equal (or really close together) but not identical, this is a bug. If two vertices are meant to be equal, but aren't identical, then they can end up with slightly different positions, depending on various factors, which would result in an error down the line (either a approximation failure, or an invalid triangle mesh being generated). It would be best to have validation checks to protect against this.

This would involve two different but similar validation checks: One [for `Sketch`](https://github.com/hannobraun/Fornjot/blob/main/crates/fj-kernel/src/validate/sketch.rs), one [for `Solid`](https://github.com/hannobraun/Fornjot/blob/main/crates/fj-kernel/src/validate/solid.rs). Those checks should use [the validation configuration](https://github.com/hannobraun/Fornjot/blob/fe28af5977590cbf4ffca158d19e3fc8b997097d/crates/fj-kernel/src/validate/mod.rs#L51-L55) to determine which vertices should be identical.

hannobraun:
Part of this issue (validation within `Solid`) has been addressed in #1695. As @A-Walrus alludes to in the commit listed above, it's probably best to hold off on the `Sketch`-related part of this issue until #1691 has been addressed.

hannobraun:
I think this issue is suited for anyone looking to start contributing to Fornjot. There is an [existing validation of that kind for `Solid`](https://github.com/hannobraun/fornjot/blob/48cf0b4dff7228349745478c48c77b2ade8ccbad/crates/fj-core/src/validate/solid.rs#L66-L129) (which was part of the original issue description), so basically all that has to be done is to [add a validation test for `Sketch` here](https://github.com/hannobraun/fornjot/blob/main/crates/fj-core/src/validate/sketch.rs), using the existing one as a template.

Also adding a unit test for the new validation check would be great, but is not required. If you're feeling ambitious, [`Shell` validation has quite a few unit tests](https://github.com/hannobraun/fornjot/blob/48cf0b4dff7228349745478c48c77b2ade8ccbad/crates/fj-core/src/validate/shell.rs#L286-L400), which show how that kind of thing could be tested.

[...]

PR #1801, opened 2023-04-27
Rework Sketches

### LINKED ISSUE #1691: Redefine relationship between `Sketch` and `Face`
Shapes in the Fornjot kernel are made up of objects (faces, edges, vertices, ...). Those objects reference each other, and the resulting object graph defines the shapes. This object graph has been simplified a lot recently (https://github.com/hannobraun/Fornjot/issues/1589), to reduce redundancy and represent shapes in a way that is as correct-by-design as possible.

The topic of this issue is one object relationship that isn't satisfactory yet, the relationship between [`Sketch`](https://docs.rs/fj-kernel/0.40.0/fj_kernel/objects/struct.Sketch.html) and [`Face`](https://docs.rs/fj-kernel/0.40.0/fj_kernel/objects/struct.Face.html). `Sketch` is a top-level object, meaning it isn't itself referenced and more-or-less represents a whole shape. `Sketch` references an arbitrary number of `Face`s which make up the `Sketch`. `Face` is also referenced by [`Shell`](https://docs.rs/fj-kernel/0.40.0/fj_kernel/objects/struct.Shell.html), which itself is referenced by [`Solid`](https://docs.rs/fj-kernel/0.40.0/fj_kernel/objects/struct.Solid.html), another top-level object.

`Face` works well in its role within the object graph of `Solid`/`Shell`. Each face is defined by the surface it's on (which could be different for any `Face` in the `Shell`), its exterior cycle, possibly interior cycles, and a color. But the same representation does not work well within a `Sketch`. The `Face`s of a `Sketch` being in different surfaces would be wrong, and actually, the surface doesn't need to be defined at all within the context of a sketch.

I've come up with the following possible solutions to this problem:

1. Add validation checks to make sure that all `Face`s in a `Sketch` are defined in the same surface. This would be problematic for the following reasons:
   - It uses validation to paper over the redundancy within the object graph. One theme of the recent cleanup has been to rather remove such redundancy, and I think the object graph has benefited a lot from that.
   - As noted above, `Sketch` doesn't need to be defined in a surface at all, and in fact I think it would be cleaner if it weren't. We only need a surface once we somehow apply this sketch to a 3D model (by sweeping it, for example).
2. Invert the relationship. Move all of `Face`'s fields, except for `surface`, to `Sketch`, and reference `Sketch` from `Face`. This solution has problem too:
   - A `Face` needs exactly one exterior cycle, while a sketch could be made up of any number of cycles that are nested in arbitrary ways.
   - Such a 1:1 relationship between `Face` and `Sketch` doesn't meet the requirements we are going to face. At some point, we'll need the capability to draw a sketch, then apply it to a face, to extrude it for example. Applying a sketch to a face in this way would would be different from referencing the sketch from the face. This mismatch would at the very least be very awkward.
3. Break the relationship. Don't reference `Face` from `Sketch`, and instead reference `Cycle`s there directly. How exactly those cycles are referenced, all in a single collection, for example, or using some structure that reflects their nesting, is an open question. In principle, I don't think you need much structure for the cycles within a `Sketch`, but how you do it influences how to later apply this sketch (or convert it) to a face.

I currently favor option 3., as it seems to have the fewest drawbacks. The details of the implementation would have to be figured out, however.

A-Walrus:
Closes #1691.
Currently it is in a semi-working state. Tests are passing, and example models work. There are a couple things I'm not sure about so would love some feedback, see comments below


crates/fj-kernel/src/algorithms/approx/sketch.rs
```diff
@@ -12,9 +12,9 @@ impl Approx for &Sketch {
 
     fn approx_with_cache(
         self,
-        tolerance: impl Into<Tolerance>,
-        cache: &mut Self::Cache,
+        _tolerance: impl Into<Tolerance>,
+        _cache: &mut Self::Cache,
     ) -> Self::Approximation {
-        self.faces().approx_with_cache(tolerance, cache)
+        todo!()
```
A-Walrus, line 18: I don't really know what this code is supposed to do...

crates/fj-kernel/src/algorithms/transform/sketch.rs
```diff
@@ -7,15 +7,10 @@ use super::{TransformCache, TransformObject};
...
+        _services: &mut Services,
+        _cache: &mut TransformCache,
     ) -> Self {
-        let faces =
-            self.faces().into_iter().cloned().map(|face| {
-                face.transform_with_cache(transform, services, cache)
-            });
-
-        Self::new(faces)
+        todo!()
```
A-Walrus, line 14: Is this even possible now that sketch is entirely 2d? We can't apply a 3D transform to a 2D sketch?

crates/fj-operations/src/lib.rs
```diff
@@ -51,16 +51,16 @@ pub trait Shape {
...
     fn compute_brep(
         &self,
         services: &mut Services,
         debug_info: &mut DebugInfo,
     ) -> Self::Brep {
         match self {
-            Self::Shape2d(shape) => {
-                shape.compute_brep(services, debug_info).faces().clone()
+            Self::Shape2d(_shape) => {
+                todo!()
```
A-Walrus, line 63: Now that sketch is completely 2D, the `Brep` of a `Shape2d` can't be a `FaceSet`

A-Walrus:
Also I turned `FaceSet` into a generic `Set<Face>`, so that `Sketch` can contain `Set<Region>`

hannobraun:
Thank you, @A-Walrus! I'll try to take a look tomorrow.

hannobraun:
I haven't had a chance to do a full review yet, but I've answered the questions. Please let me know, if there's anything else that's not clear!

crates/fj-kernel/src/algorithms/transform/sketch.rs
A-Walrus: I removed `TransformObject` for `Sketch`, without touching `fj-operations`/`fj` and everything seems to work fine...
Is there anything specific I should check/change?

crates/fj-kernel/src/objects/full/face.rs
```diff
@@ -35,9 +35,7 @@ use crate::{
 #[derive(Clone, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
 pub struct Face {
     surface: Handle<Surface>,
-    exterior: Handle<Cycle>,
-    interiors: Vec<Handle<Cycle>>,
-    color: Option<Color>,
+    region: Region,
```
A-Walrus, line 38: Should this be a `Region` or `Handle<Region>`?

crates/fj-operations/src/lib.rs
A-Walrus: Allowing the `Brep` to be a sketch, we would run into a problem where in:
 https://github.com/hannobraun/Fornjot/blob/a2579f6ee99793a9bdc12b499454dd8d9667fec1/crates/fj-operations/src/transform.rs#L10-L20
We need to be able to transform the `Brep` which we can't do to a `Sketch`.
I think for now I'll implement it using the xy-plane :+1:

crates/fj-kernel/src/algorithms/transform/sketch.rs
hannobraun: Oh yeah, that's because none of the `Shape::Brep` associated type are actually set to `Sketch`. Since you decided to convert it to 3D using a default plane, no changes will be required.

crates/fj-kernel/src/objects/full/face.rs
hannobraun: That question is basically the same as asking, is `Region` an object (i.e. stored in centralized storage and referred to via `Handle`) or is it just a normal struct. It should be an object, if it needs an identity ([documented here](https://docs.rs/fj-kernel/0.44.0/fj_kernel/storage/struct.Handle.html#equality-and-identity)), which can be relevant for validation (see https://github.com/hannobraun/Fornjot/issues/1689, for example).

The the answer will probably become obvious at some point. For now, I don't know, but I can't think of a reason why it needs to be an object. I think that means it would be best to make it a normal struct, as that's less complex. If it needs to become an object at some point, we can make that happen then.

If it's not an object, that means it should probably live in [`geometry`](https://docs.rs/fj-kernel/0.44.0/fj_kernel/geometry/index.html), rather than within the `objects` module.

Good catch, by the way! I hadn't thought about this before, and had just assumed that it needs to be an object.

hannobraun:
Hey, @A-Walrus! I've recently decided to reduce the scope of this project (see [A New Direction](https://www.fornjot.app/blog/a-new-direction/) for context). As a result, parts of the code changed here [have been removed](https://github.com/hannobraun/fornjot/pull/1822). Namely, the `fj-operations` crate is no longer part of this repository.

As far as I'm aware, all open questions have been addressed, so I guess aside from a rebase to fix the merge conflixts, this should be ready for review?

hannobraun:
Superseded by #1828. Closing.

PR #1828, opened 2023-05-15
Rework Sketches v2

A-Walrus:
Version 2 of #1801, this version is rebased and much simpler. Not sure how to test that this is working with the new changes? Tests are passing, but other than that I don't know what to check...

### DIFF
--- a/crates/fj-kernel/src/geometry/region.rs
+++ b/crates/fj-kernel/src/geometry/region.rs
@@ -8,7 +8,7 @@ use crate::{
     storage::Handle,
 };
 
-/// A single, continues 2d region, may contain holes. Once applied to a
+/// A single, continuous 2d region, may contain holes. Once applied to a
 /// [`Surface`] becomes a [`Face`]
 ///
 /// Interior cycles must have the opposite winding of the exterior cycle,


crates/fj-kernel/src/algorithms/approx/sketch.rs
```diff
@@ -12,9 +12,9 @@ impl Approx for &Sketch {
 
     fn approx_with_cache(
         self,
-        tolerance: impl Into<Tolerance>,
-        cache: &mut Self::Cache,
+        _tolerance: impl Into<Tolerance>,
+        _cache: &mut Self::Cache,
     ) -> Self::Approximation {
-        self.faces().approx_with_cache(tolerance, cache)
+        todo!()
```
A-Walrus, line 18: Left this as a `todo!()` for now as I'm not aware that we need to approximate `Sketch`es directly _yet_, we always first turn them into faces/shells and then approximate?

hannobraun: Yeah, fair enough. We can fill this in once required :+1:

crates/fj-kernel/src/geometry/region.rs
```diff
@@ -0,0 +1,77 @@
...
+use fj_interop::mesh::Color;
+
+use crate::{
+    objects::{Cycle, Face, Surface},
+    operations::Insert,
+    services::Services,
+    storage::Handle,
+};
+
+/// A single, continues 2d region, may contain holes. Once applied to a
```
hannobraun, line 11: Typo:
```suggestion
/// A single, continuous 2d region, may contain holes. Once applied to a
```

hannobraun (approved):
Thank you, @A-Walrus, looks great!

I found one typo and left a comment with a suggestion. I can probably apply that myself, after submitting this review.

> Not sure how to test that this is working with the new changes? Tests are passing, but other than that I don't know what to check...

Yeah, sorry. We temporarily lost the example models, and the automated testing associated with them. I'll re-add those soon. For now, I'm happy to merge anything that compiles and looks reasonable.

PR #1933, opened 2023-07-10
Update dependencies

hannobraun:
Close #1923
Close #1924
Close #1925
Close #1926
Close #1927
Close #1928
Close #1929
Close #1930
Close #1931
Close #1932


AlexDaniel:
FYI you can use https://github.com/renovatebot/renovate to avoid PR spam from dependabot. Renovate can be configured to create a single PR with all of the updates, it's a lot less annoying.

hannobraun:
Thanks for letting me know, @AlexDaniel!

Dependabot has this feature too. See [the latest PR](https://github.com/hannobraun/fornjot/pull/2001), for example.

I recently became aware of Renovate, and I would be tempted to switch, if I hadn't discovered this Dependabot feature around the same time.

AlexDaniel:
@hannobraun oh wow, they finally did it! Looks like the feature became available in dependabot just two months ago. Nice, good to know! 😇

hannobraun:
Yeah, it seems pretty recent. It's currently in public beta, I think.

It's still not doing a `cargo update` though. Currently I'm pushing a commit to the Dependabot PR to do that manually. Not sure whether it's better to wait for them to grow that feature, or go through the trouble to switch to another solution :smile:

PR #1954, opened 2023-07-21
Remove out-of-date note

A-Walrus:
We do validate that a shell is watertight, see: https://github.com/hannobraun/fornjot/blob/40b364814a6f9f0c85f508e637941ee0cc2bd0eb/crates/fj-core/src/validate/shell.rs#L24C40-L24C40


hannobraun (approved):
Thank you, @A-Walrus, good catch!

PR #1956, opened 2023-07-22
Remove fj.toml file

TobiasJacob:
It seems to me that fj.toml file is not used anymore. If this is correct, you can approve this PR, otherwise just reject it.


hannobraun (approved):
Thank you, @TobiasJacob, that is correct. Good catch!

(I missed this in https://github.com/hannobraun/fornjot/pull/1822.)

Issue #1935, opened 2023-07-11
Validate face orientation of `Solid`s


hannobraun:
The orientation of a face is defined by the winding of its exterior cycle. On the front side of the face, the exterior cycle is wound counter-clockwise. It is possible to create `Solid`s where some or all of the faces face inwards, which would lead to invalid shading when displaying the `Solid`, and possibly invalid meshes when exporting it to an external file format.

This should be caught in a validation check, which makes sure that all faces of every solid point outwards.

A-Walrus:
Some questions:

- Should we be checking that all faces of a `Shell` have the same orientation (which can be either inwards or outwards)?
- If a `Solid` has a cavity, what should be the orientation of the faces of the cavity?

hannobraun:
> Should we be checking that all faces of a `Shell` have the same orientation (which can be either inwards or outwards)?

Yeah, that makes sense! A mixed-orientation `Shell` would definitely be invalid (even though that's not specified anywhere right now), and having that check in place would then make the `Solid` check easier

> If a `Solid` has a cavity, what should be the orientation of the faces of the cavity?

The cavity is "outside" of the `Solid`. So if you're sitting inside of the cavity looking at the solid, you should see counter-clockwise faces.

hannobraun:
I've been looking into coordinate systems and cycle winding while working on https://github.com/hannobraun/fornjot/issues/2098, and I found out that the situation is not as simple as I presented it in the issue description above. I said that the orientation of a face is defined by the winding of its exterior cycle, and that is simply wrong. The orientation of a face is defined by the *combination of* the winding and the handedness of the surface coordinate system.

To summarize:

- When looking at a face **from the front** (i.e. looking at a shell from the outside):
  - Winding must be **counter-clockwise** and surface coordinate system must be **right-handed**, or
  - winding must be **clockwise** and surface coordinate system must be **left-handed**.
- When looking at a face **from the back** (i.e. looking at a shell from inside the shell):
  - Winding must be **counter-clockwise** and surface coordinate system must be **left-handed**, or
  - winding must be **clockwise** and surface coordinate system must be **right-handed**.

[...]

Issue #1937, opened 2023-07-12
Lift limitation, that coincident `HalfEdge`s must be congruent


hannobraun:
The approximation code makes the assumption, that coincident `HalfEdge`s are always congruent. This [requirement is documented](https://github.com/hannobraun/fornjot/blob/3bcb450914f2af650b0b9e27773f66d5fdc2a39b/crates/fj-core/src/objects/kinds/edge.rs#L27-L30), as is [the reason for it](https://github.com/hannobraun/fornjot/blob/53a49ea00b976ac580353bd07a0f9128caa5b116/crates/fj-core/src/algorithms/approx/edge.rs#L49-L89).

I'd like to implement an operation that can split the faces of a shell, which will be useful for further manipulation of the shell. In the presence of such an operation, maintaining this limitation becomes wholly impractical. Hence, I think it's time to remove it. This will require changes to the object graph.

I'm sure an appropriate design will reveal itself as I'm working on it, but for now I'm thinking we need a new object to represent a curve in global space. That would then be referenced by `HalfEdge`, and `GlobalEdge` would become redundant and could be removed. But we'll see how it shakes out. (Idea on nomenclature: Maybe the current `Curve` can become `Path` or `GlobalPath`, then the new object can be called `Curve`.)

TobiasJacob:
Hi, I'm still  new to this repo, but I think I understand the problem.

Approximating / Meshing a `Solid` requries knowledge of which edges are coincident, such that no gaps between connected faces appear due to approximation errors.

I wonder if `Faces` in a `FaceSet` (named FaceGroup in Ian Strouds book) should have to be continuous. This means that neigbouring faces in a `FaceSet` always have a joined `HalfEdge`. Then, a `FaceSet` could store the information for the Global Edge / Curve object.

I think the idea of renaming Curve to SurfacePath and calling the new object Curve makes sense. I think it helps to try to stick to the naming conventions from Ian Stroud - Boundary Representation Modelling Techniques.

![image](https://github.com/hannobraun/fornjot/assets/11909751/6d465c33-64d0-4b94-ad98-da4b17d0bc66)

hannobraun:
Thank you for the feedback, @TobiasJacob!


> Approximating / Meshing a `Solid` requries knowledge of which edges are coincident, such that no gaps between connected faces appear due to approximation errors.

[...]

PR #1968, opened 2023-07-29
Validate shell orientations

A-Walrus:
Addresses part of #1935.

Currently bugged, so marked as draft.
The problem is how to check whether two half-edges are reverse of one another:
- Make sure that start vertexes are different (fails on circles)
- Make sure that boundarys are reverse of one another (fails when different paths)
- Sample the edges?

Any help appreciated, thanks!

### DIFF
--- a/crates/fj-core/src/validate/shell.rs
+++ b/crates/fj-core/src/validate/shell.rs
@@ -371,7 +371,7 @@ impl ShellValidationError {
         for (_, halfs) in global_to_half {
             if let (Some(a), Some(b)) = (halfs.get(0), halfs.get(1)) {
                 // Check if a is reverse of b
-                if a.start_vertex().id() == b.start_vertex().id() {
+                if a.boundary().reverse() != b.boundary() {
                     errors.push(Self::MixedOrientations.into());
                     dbg!(a, b);
                     return;
--- a/crates/fj-core/src/validate/shell.rs
+++ b/crates/fj-core/src/validate/shell.rs
@@ -20,6 +20,7 @@ impl Validate for Shell {
         ShellValidationError::validate_curve_coordinates(self, config, errors);
         ShellValidationError::validate_edges_coincident(self, config, errors);
         ShellValidationError::validate_watertight(self, config, errors);
+        ShellValidationError::validate_same_orientation(self, errors);
     }
 }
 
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you for the pull request, @A-Walrus! And sorry for taking a few days to get to it. I had other things on my mind over the weekend :smile: 



> The problem is how to check whether two half-edges are reverse of one another:
> 
> * Make sure that start vertexes are different (fails on circles)
> * Make sure that boundarys are reverse of one another (fails when different paths)
> * Sample the edges?

I think checking that the boundaries are reverse should be the answer here. That requires the coordinate systems of all local path that refer to the same curve to match each other. Which I think is a reasonable requirement, exactly because we need to address cases like this.

We already have the `Shell` validation check that checks edge coincidence, but it's written in such a way, that it doesn't require the path coordinate systems to match. Maybe a good next step would be to rewrite that check to require just that, see how much fails, and figure out what it would take to make it work.

hannobraun:
To follow up on my previous comment, I've opened https://github.com/hannobraun/fornjot/issues/1973. This directly affects my work on https://github.com/hannobraun/fornjot/issues/1937 (as in, the code I'm currently writing is unlikely to work due to this), so I'm looking into it now.

hannobraun:
@A-Walrus I've addressed https://github.com/hannobraun/fornjot/issues/1973 in https://github.com/hannobraun/fornjot/pull/1982. Unless I'm missing something, that means further work on this pull request should be un-blocked now!

crates/fj-core/src/validate/shell.rs
```diff
@@ -324,24 +329,56 @@ impl ShellValidationError {
...
+        if global_edge_to_faces.iter().any(|(_, c)| *c != 2) {
             errors.push(Self::NotWatertight.into())
         }
     }
+
+    fn validate_same_orientation(
+        shell: &Shell,
+        errors: &mut Vec<ValidationError>,
+    ) {
+        let mut global_to_half: HashMap<ObjectId, Vec<_>> = HashMap::new();
```
hannobraun, line 355: For some additional type safety, it should be possible to replace `ObjectdId` with `HandleWrapper<GlobalEdge>` here. This was not possible before, due to a horribly shameful workaround (:grin:), but that workaround was removed in https://github.com/hannobraun/fornjot/pull/1951.

This is just a note, not a change request. I don't think it would make a difference here.

hannobraun (approved):
Thank you, @A-Walrus, looks good!

One thing to note is that the new validation check relies on `GlobalEdge`, which is on its way to being removed. There is already the new `Curve`, which fulfills the same purpose but is more flexible.

As part of my work on https://github.com/hannobraun/fornjot/issues/1937, I've already amended the other validation checks to also use `Curve`, in preparation of removing `GlobalEdge` at a later point.

All of that is no reason not to merge this pull request. I'm just providing some context on how the code being changed here is currently developing.

PR #2049, opened 2023-10-09
Update dependencies

hannobraun:
Follow-up to https://github.com/hannobraun/fornjot/pull/2047. No idea what was going on there, but hopefully a new pull request will take care of it.


hannobraun:
And, Dependabot opened its own new pull request in the meantime: https://github.com/hannobraun/fornjot/pull/2048

Going forward with that one.

PR #2087, opened 2023-11-11
Inling in `fj-math` and general optimizations

lxrst:
Mostly just inlined a bunch of `fj-math` stuff. Also set the `dev` and `release` profiles to something that should be more suitable.

I'd like to do whatever I can to help out. Is there anything in particular I should work on?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ path = "crates/fj-window"
 opt-level = 3
 
 [profile.dev]
-opt-level = 0
+opt-level = 1
 lto = false
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,3 +71,15 @@ path = "crates/fj-viewer"
 [workspace.dependencies.fj-window]
 version = "0.47.0"
 path = "crates/fj-window"
+
+[profile.dev.package.'*']
+opt-level = 3
+
+[profile.dev]
+opt-level = 0
+lto = false
+
+[profile.release]
+opt-level = 3
+lto = true
+strip = true
--- a/crates/fj-math/src/scalar.rs
+++ b/crates/fj-math/src/scalar.rs
@@ -1,5 +1,5 @@
 use std::{
-    cmp,
+    cmp::{self, Ordering},
     f64::consts::{PI, TAU},
     fmt,
     hash::Hash,
@@ -42,6 +42,7 @@ impl Scalar {
     /// # Panics
     ///
     /// Panics, if `scalar` is NaN.
+    #[inline(always)]
     pub fn from_f64(scalar: f64) -> Self {
         if scalar.is_nan() {
             panic!("Invalid scalar value: {scalar}");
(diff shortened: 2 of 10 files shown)


hannobraun:
Thank you for the pull request, @lxrst! I currently don't have time to look into it, but I will do so next week.

> I'd like to do whatever I can to help out. Is there anything in particular I should work on?

Unfortunately, I don't have a good answer right now. My own time is  more limited these days, and contributions have been low for a while, so I'm not as on top of opening suitable issues as I used to be. The things that are currently on top of my mind are pretty deep architectural issues that are probably not suited for a new contributor, at least not without me making the significant effort of thinking it all through and writing it up.

I'll think on it!

crates/fj-math/src/scalar.rs
```diff
@@ -89,26 +97,23 @@ impl Scalar {
...
+            Ordering::Equal => Sign::Zero,
         }
-        if self.is_positive() {
-            return Sign::Positive;
-        }
-        if self.is_zero() {
-            return Sign::Zero;
-        }
-
-        unreachable!("Sign is neither negative, nor positive, nor zero.")
```
hannobraun, line 103: This is a nice change! If it were in its own commit, I'd merge it right now!

Cargo.toml
```diff
@@ -71,3 +71,15 @@ path = "crates/fj-viewer"
 [workspace.dependencies.fj-window]
 version = "0.47.0"
 path = "crates/fj-window"
+
+[profile.dev.package.'*']
+opt-level = 3
```
hannobraun, line 76: I'm very doubtful that this provides a benefit. I think it will increase the compile times of everything in `crates/`, whenever you run an example model (which is one of the main forms of testing when developing in this repository).

Cargo.toml
```diff
@@ -71,3 +71,15 @@ path = "crates/fj-viewer"
...
+
+[profile.dev]
+opt-level = 1
+lto = false
+
+[profile.release]
+opt-level = 3
+lto = true
+strip = true
```
hannobraun, line 85: Why did you choose those specific settings? They look very deliberate, but I can't tell why they might be better than others. I'm specifically wary of the `dev` settings, as those have the potential to make iteration times during development worse.

hannobraun (commented):
Hey @lxrst, I had a chance to look at the pull request now. Thanks again!

I'm not sure what to make of it, though. Some general comments:

1. There is a cost to all those `#[inline]`s, in terms of visual noise that makes the code harder to read. I'm not opposed to adding them, but I'd like to know what benefit we're getting in return. Is there an objective measurement that shows such a benefit?
2. Similarly, does inlining those variables actually provide a benefit? I'd be *very* surprised, if there was any performance benefit to it. I'd be willing to merge something like that were it makes the code more readable (which some of those changes might do and others definitely don't).

I've also left some more specific comments in the code.

Cargo.toml
hannobraun: Ah, thanks for clarifying. I must have thought it it applied to dependencies in general, regardless of whether they're in the workspace or not. That's much better.



> I think I added this to help with `egui`?

It's worth noting then, that `egui` is no longer a dependency.

hannobraun:
Thank you for getting back to me, @lxrst!

Regarding profile configuration and inlining, you explained _what_ these change do, in general. Not what they do specifically in this case, for Fornjot, or _why_ you made them.

As a general rule, I'm happy to merge anything that's an improvement. But adding more lines of code, by itself, is the opposite. It's more for new contributors to understand, and more for me to maintain.

More code has to create a benefit to be worth merging. And while I'm aware that inlining, for example, can have a positive impact on performance, it is not clear to me that the specific changes in this pull request have a positive impact on Fornjot.

Can a positive impact be measured? Maybe by adding benchmarks to the repository, or just by doing some quick tests locally? Then I'm happy to merge it. But I'm wary of merging these kinds of changes without such evidence. That's just cargo-cult programming.

hannobraun:
I would recommend putting that into a separate pull request. That makes it easier to understand what's happening, by looking at what pull requests got merged.

The bigger question is, if it's worth doing at this point. So far, performance hasn't been a priority for me, in favor of getting things to actually work first. I expect that not to change for a while. And once it changes, I expect the focus to be on big, architectural optimizations for a while.

If you still want to work on it, I'm happy to merge anything that looks reasonable. But I won't be able to provide much guidance (due to not thinking much about the topic so far). It's also worth nothing that any benchmarks that get added would be a rather isolated thing in this repository (due to the lack of focus on performance), and that kind of thing is always at danger of getting removed again, if it ever gets in the way.

So in summary, I'm not saying no, but I also provide context. I don't want you to get the feeling that you're wasting your time.

Issue #2133, opened 2023-12-12
Validate object ownership


hannobraun:
There are some objects in the graph that are essentially owned by the objects that reference them, meaning that there's typically exactly one object that references them. I believe that for these objects, it's always a mistake to reference them from multiple other objects, within the context of a valid `Sketch` or `Solid`.

I believe that the following should be true, for a valid `Sketch` or `Solid`:

- [ ] Each `HalfEdge` is referenced by exactly one `Cycle`.
- [ ] Each `Cycle` is referenced by exactly one `Region`.
- [ ] Each `Region` is referenced by exactly one `Sketch` or `Face`. (edit: not necessary for `Sketch`; see https://github.com/hannobraun/fornjot/pull/2144#pullrequestreview-1817856036)
- [ ] Each `Face` is referenced by exactly one `Shell`.

I believe that these checks won't be hard to implement, so this might be suitable for new contributors:

- Validation checks for `Sketch` and `Solid` need to be placed in [`validate/sketch.rs`](https://github.com/hannobraun/fornjot/blob/main/crates/fj-core/src/validate/sketch.rs) and [`validate/solid.rs`](https://github.com/hannobraun/fornjot/blob/main/crates/fj-core/src/validate/solid.rs) respectively.
- Each item in the list above should go into a separate check, for clarity. You can check out the other files in the `validate` module for examples on how these checks are typically structured.
- If you're up for it, it would be ideal if you included a unit test with each check. There are many examples of those in the various `validate` sub-modules.
- The APIs of [`Sketch`](https://github.com/hannobraun/fornjot/blob/main/crates/fj-core/src/objects/kinds/sketch.rs) and [`Solid`](https://github.com/hannobraun/fornjot/blob/main/crates/fj-core/src/objects/kinds/solid.rs) themselves are quite straight-forward. You can easily access the objects they reference, use those objects to access further objects, and so forth.
- For a start, it would be enough to just iterate over all objects of the respective type and remember which objects they reference. We can leave smarter and better performing approaches for future work.

If you're interested in working on this, please don't hesitate to ask any questions you might have!

nathan-folsom:
Hey @hannobraun, I just did an initial pass at implementing the check for HalfEdges in #2144. Am I on the right track?

Also a question about the errors themselves: Are there any details that would be useful to surface about the circumstances of the failed validation, e.g. number of times an object was references, what it was referenced by? Or is it enough to just report that an object was referenced more than once?

[...]

PR #2144, opened 2023-12-17
Object Reference Validation #2133

nathan-folsom:
Add validation of object references in Sketches and Solids as outlined in #2133

### DIFF
--- a/crates/fj-core/src/services/service.rs
+++ b/crates/fj-core/src/services/service.rs
@@ -1,4 +1,4 @@
-use std::ops::{Deref, DerefMut};
+use std::ops::Deref;
 
 /// A service that controls access to some state
 ///
@@ -63,7 +63,7 @@ impl<S: State> Deref for Service<S> {
 }
 
 #[cfg(test)]
-impl<S: State> DerefMut for Service<S> {
+impl<S: State> std::ops::DerefMut for Service<S> {
     fn deref_mut(&mut self) -> &mut Self::Target {
         &mut self.state
     }
--- a/crates/fj-core/src/services/service.rs
+++ b/crates/fj-core/src/services/service.rs
@@ -62,6 +62,7 @@ impl<S: State> Deref for Service<S> {
     }
 }
 
+#[cfg(test)]
 impl<S: State> DerefMut for Service<S> {
     fn deref_mut(&mut self) -> &mut Self::Target {
         &mut self.state
--- a/crates/fj-core/src/services/service.rs
+++ b/crates/fj-core/src/services/service.rs
@@ -1,4 +1,4 @@
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
 
 /// A service that controls access to some state
 ///
@@ -59,6 +59,12 @@ impl<S: State> Deref for Service<S> {
 
     fn deref(&self) -> &Self::Target {
         &self.state
+    }
+}
+
+impl<S: State> DerefMut for Service<S> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.state
     }
 }
 
--- a/crates/fj-core/src/validate/sketch.rs
+++ b/crates/fj-core/src/validate/sketch.rs
@@ -41,21 +41,15 @@ impl SketchValidationError {
 
         sketch.regions().iter().for_each(|r| {
             r.all_cycles().for_each(|c| {
-                referenced_cycles.insert(c.clone(), {
-                    if let Some(count) = referenced_cycles.get(c) {
-                        count + 1
-                    } else {
-                        1
-                    }
-                });
+                referenced_cycles
+                    .entry(c.clone())
+                    .and_modify(|count| *count += 1)
+                    .or_insert(1);
                 c.half_edges().into_iter().for_each(|e| {
-                    referenced_edges.insert(e.clone(), {
-                        if let Some(count) = referenced_edges.get(e) {
-                            count + 1
-                        } else {
-                            1
-                        }
-                    });
+                    referenced_edges
+                        .entry(e.clone())
+                        .and_modify(|count| *count += 1)
+                        .or_insert(1);
                 })
             })
         });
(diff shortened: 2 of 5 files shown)


crates/fj-core/src/validate/sketch.rs
```diff
@@ -1,12 +1,117 @@
...
+    fn check_obect_references(
+        sketch: &Sketch,
+        _config: &ValidationConfig,
+        errors: &mut Vec<ValidationError>,
+    ) {
+        let mut referenced_edges =
+            std::collections::HashMap::<Handle<HalfEdge>, i32>::new();
+
+        // Do we care about how many times each edge is used, or should we just return as soon as
+        // we find one that is used more than once?
```
hannobraun, line 39: Experience has shown that early-aborting validation checks is generally a bad idea (which is why they're done in the background through a service, instead of using the normal Rust `Result`-based error handling).

With early abort, you often end up with not enough information to figure out what's going wrong.

crates/fj-core/src/validate/sketch.rs
```diff
@@ -32,13 +32,22 @@ impl SketchValidationError {
...
-        // we find one that is used more than once?
         sketch.regions().iter().for_each(|r| {
             r.all_cycles().for_each(|c| {
+                referenced_cycles.insert(c.clone(), {
+                    if let Some(count) = referenced_cycles.get(c) {
+                        count + 1
+                    } else {
+                        1
+                    }
+                });
```
hannobraun, line 50: I still haven't done a real review (as I said over in https://github.com/hannobraun/fornjot/issues/2133, I'm on vacation right now), but this jumped out to me.

To be clear, this is fine, I would merge it as-is. I just wanted to mention, for the sake of education, that this looks like the perfect use case for the [entry API](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry). Something like this should work (not compiled or tested):

```suggestion
                *referenced_cycles.entry(c.clone()).or_insert(0) += 1;
```

Feel free to take this suggestion, or not, as you see fit.

nathan-folsom:
Ah yes, that looks like the right way to do this. I will add it to the cleanup list! I’m generally familiar with rust syntax, but still learning the finer details so happy to consider any suggestions you might have.On Dec 18, 2023, at 6:34 AM, Hanno Braun ***@***.***> wrote:﻿
@hannobraun commented on this pull request.



In crates/fj-core/src/validate/sketch.rs:
> +                referenced_cycles.insert(c.clone(), {
+                    if let Some(count) = referenced_cycles.get(c) {
+                        count + 1
+                    } else {
+                        1
+                    }
+                });

I still haven't done a real review (as I said over in #2133, I'm on vacation right now), but this jumped out to me.
To be clear, this is fine, I would merge it as-is. I just wanted to mention, for the sake of education, that this looks like the perfect use case for the entry API. Something like this should work (not compiled or tested):
⬇️ Suggested change
-                referenced_cycles.insert(c.clone(), {
-                    if let Some(count) = referenced_cycles.get(c) {
-                        count + 1
-                    } else {
-                        1
-                    }
-                });
+                *referenced_cycles.entry(c.clone()).or_insert(0) += 1;


Feel free to take this suggestion, or not, as you see fit.

nathan-folsom:
@hannobraun Since you last looked, did some cleanup and included referenced + referencing objects in error messages. Feels merge-able to me, but let me know what you think when you have the chance!

hannobraun:
Thank you for the updates, @nathan-folsom! I am now confident that I won't review this before January. In addition to being on vacation, I've also picked up a cold on Christmas. Little motivation to do anything but watch movies right now :smile: 

Sorry for the delay. Usually I handle these within a few days, at most.

nathan-folsom:
No problem, hope you're feeling better!

hannobraun:
I am, thank you! Getting closer to where I can review this.

crates/fj-core/src/validate/sketch.rs
```diff
@@ -1,12 +1,140 @@
...
+/// [`Sketch`] validation failed
+#[derive(Clone, Debug, thiserror::Error)]
+pub enum SketchValidationError {
+    /// Object within sketch referenced by more than one other object
+    #[error("Object within sketch referenced by more than one other Object")]
+    MultipleReferences(#[from] ReferenceCountError),
+}
+
+impl SketchValidationError {
+    fn check_obect_references(
```
hannobraun, line 27:
```suggestion
    fn check_object_references(
```

(Also requires changing the caller, of course, so just clicking "accept" on this suggestion won't be enough.)

crates/fj-core/src/services/service.rs
```diff
@@ -62,6 +62,12 @@ impl<S: State> Deref for Service<S> {
     }
 }
 
+impl<S: State> DerefMut for Service<S> {
```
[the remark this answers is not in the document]
hannobraun: It feels wrong, but I don't have a better idea right now. Since it's only used in tests, we should slap a `#[cfg(test)]` on there, to make sure it's not misused by other code.

hannobraun (changes requested):
Hey @nathan-folsom, thank you again for the pull request. And I'm very sorry it took me so long to get to it. There was a confluence of factors, but I'm finally back in the game now.

I've left some comments. The only problem I'd like to see addressed, is the `#[cfg(test)]` for the `DerefMut` implementation, since that could be error-prone, if left as-is. I'd be happy to merge the pull request with the typo remaining (it can be fixed later, and I'd rather get this merged as soon as practical).

One note, since this confused me for a bit: My original list in https://github.com/hannobraun/fornjot/issues/2133 says we need to validate (on the `Sketch` level) that each `Region` is referenced by exactly one `Sketch`. This pull request doesn't do that, which is actually correct! There's only one `Sketch`, and it references its regions via one `ObjectSet`, which already guarantees uniqueness. So all good!

nathan-folsom:
@hannobraun Done and done!

hannobraun (approved):
Thank you, @nathan-folsom, looks awesome!

Issue #2060, opened 2023-10-19
Make background validation performed by `Services` configurable


hannobraun:
Validation uses parameters defined in `ValidationConfig` to determine whether close-by objects are considered coincident or not. However, this is not exposed through the `Services` API, which is the main way of using validation.

It should be possible to override the configuration used by `Services` in some way.

hannobraun:
I think this might be a good first issue for anyone interested in getting involved. Here's some more explanation and guidance on how to get this done:

- [`Services`](https://github.com/hannobraun/fornjot/blob/48cf0b4dff7228349745478c48c77b2ade8ccbad/crates/fj-core/src/services/mod.rs#L20-L31) is currently the entry point to the `fj-core` crate.
- You can see examples of its use in the example models ([[1]](https://github.com/hannobraun/fornjot/blob/48cf0b4dff7228349745478c48c77b2ade8ccbad/models/cuboid/src/main.rs#L4-L5), [[2]](https://github.com/hannobraun/fornjot/blob/48cf0b4dff7228349745478c48c77b2ade8ccbad/models/cuboid/src/lib.rs#L12-L31), for example) or pretty much any unit test within `fj-core` ([like this one](https://github.com/hannobraun/fornjot/blob/48cf0b4dff7228349745478c48c77b2ade8ccbad/crates/fj-core/src/validate/shell.rs#L299-L338)).
- One service within `Services` is the [validation service](https://github.com/hannobraun/fornjot/blob/main/crates/fj-core/src/services/validation.rs), which [uses the default `ValidationConfig` by calling `Validate::validate`](https://github.com/hannobraun/fornjot/blob/48cf0b4dff7228349745478c48c77b2ade8ccbad/crates/fj-core/src/services/validation.rs#L55).
- So to address this issue, one would have to optionally accept a `ValidationConfig` in `Services` (maybe through a second constructor), pass that config into the validation service, and use it there via [`Validate::validate_with_config`](https://github.com/hannobraun/fornjot/blob/48cf0b4dff7228349745478c48c77b2ade8ccbad/crates/fj-core/src/validate/mod.rs#L128-L133).

If anyone still has questions, please don't hesitate to ask!

hannobraun:
Addressed in https://github.com/hannobraun/fornjot/pull/2150.

PR #2150, opened 2023-12-23
 Make background validation performed by Services configurable #2060

brungardtdb:
Added functionality to pass in a predefined validation configuration into Services so that it can be used in the Validation Service as outlined in #2060.


hannobraun:
Thank you for this pull request, @brungardtdb! I'm currently on vacation (for the rest of the year). I might take a look at this next week, when I get a chance, or at the latest in January, when I'm back to regular work.

crates/fj-core/src/services/mod.rs
```diff
@@ -42,6 +42,18 @@ impl Services {
         }
     }
 
+    /// Construct an instance of `Services` with a pre-defined configuration for the validation service
+    pub fn with_validation_config(config: &ValidationConfig) -> Self {
```
hannobraun, line 46: Seems a bit misleading to take a `&ValidationConfig` that is then copied inside. It would be better to directly take `ValidationConfig`.

But this is not a blocker. Happy to merge as-is!

crates/fj-core/src/services/validation.rs
```diff
@@ -3,16 +3,34 @@ use std::{collections::BTreeMap, error::Error, thread};
...
 
 use super::State;
 
 /// Errors that occurred while validating the objects inserted into the stores
-#[derive(Default)]
 pub struct Validation {
     /// All unhandled validation errors
     pub errors: BTreeMap<ObjectId, ValidationError>,
+    /// Optional validation config
+    config: Option<ValidationConfig>,
```
hannobraun, line 16: It's probably simpler to instead have `config: ValidationConfig`, and initialize that with its default value, if not provided.

But this would be a nice-to-have. Perfectly fine to merge this as-is.

crates/fj-core/src/services/service.rs
```diff
@@ -62,6 +62,12 @@ impl<S: State> Deref for Service<S> {
     }
 }
 
+impl<S: State> DerefMut for Service<S> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.state
+    }
+}
```
hannobraun, line 69: I think this is used by the new `Services` constructor, which in itself seems fine. But this provides direct access to service state to *all* code, which seems error-prone.

I think there are two better ways to do this:

1. Define a command for updating the validation config (see the existing `ValidationCommand`), make the execution of that command emit `ValidationEvent::ConfigurationDefined`, and use `Service::execute` to update the validation.
2. In the new `Services` constructor, instead of using `Service::<Validation>::default()`, initialize the validation service with `Service::new`, which can be provided with a `Validation` that has the configuration set properly.

Solution 1 is more verbose, and probably more flexible than we need. So I'd prefer solution 2. What do you think?

hannobraun (changes requested):
Thank you again for the pull request, @brungardtdb! And I'm very sorry that it took me so long to review it. There was a confluence of factors that caused this. Usually, things around here work much more quickly!

I've left a few comments. Most are just me nit-picking, and fine to leave as-is (as noted in the respective comment). The only thing I'd like to see changed before merging, is the `DerefMut` implementation.

brungardtdb:
> And I'm very sorry that it took me so long to review it. There was a confluence of factors that caused this. Usually, things around here work much more quickly!

No worries! I think I submitted this PR over the holidays, so I wasn't expecting to hear back right away. :smile:

crates/fj-core/src/services/service.rs
brungardtdb: I think I agree with you, solution 1 seems like it might be overkill, so I will try my hand at solution 2 and see how that goes. :+1:

crates/fj-core/src/services/mod.rs
brungardtdb: This has been addressed.

crates/fj-core/src/services/validation.rs
brungardtdb: This has also been addressed.

crates/fj-core/src/services/service.rs
brungardtdb: @hannobraun , I believe I have managed to implement all of the changes that you had suggested. Please take a look and let me know if this is what you were going for! :smile:

hannobraun (approved):
Awesome, @brungardtdb! Thank you!

brungardtdb:
@hannobraun, thank you!

Issue #2223, opened 2024-02-19
Upgrade to `threemf` 0.5


hannobraun:
This requires manual intervention. See https://github.com/hannobraun/fornjot/pull/2222.

hannobraun:
Addressed by @IamTheCarl in https://github.com/hannobraun/fornjot/pull/2248.

PR #2248, opened 2024-03-02
Update ThreeMF to 0.5.0

IamTheCarl:
Comically minimum effort. Completes #2223 .

I intend to make some bigger changes later but I figured this PR would be an easier starting point.

For my next PR I'd like to adapt the `export_3mf`, `export_stl`, and `export_obj` functions to accept anything implementing `std::io::Write` instead of a file path, add documentation comments, and then make them public.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2434,21 +2434,12 @@ checksum = "43d84d1d7a6ac92673717f9f6d1518374ef257669c24ebc5ac25d5033828be58"
 
 [[package]]
 name = "quick-xml"
-version = "0.27.1"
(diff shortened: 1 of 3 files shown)


hannobraun (approved):
Thank you, @IamTheCarl, looks great!

> Comically minimum effort.

Perfect! Every improvement is worth merging.

> For my next PR I'd like to adapt the `export_3mf`, `export_stl`, and `export_obj` functions to accept anything implementing `std::io::Write` instead of a file path, add documentation comments, and then make them public.

Sounds great!

PR #2252, opened 2024-03-04
Generalize export functions and make them public

IamTheCarl:
Exposes export functions for specific mesh formats and generalizes them to write to objects that implement Write (except for 3MF which needs the target to implement both Write and Seek).

### DIFF
--- a/crates/fj-export/src/lib.rs
+++ b/crates/fj-export/src/lib.rs
@@ -8,7 +8,11 @@
 //!
 //! [Fornjot]: https://www.fornjot.app/
 
-use std::{fs::File, path::Path};
+use std::{
+    fs::File,
+    io::{Seek, Write},
+    path::Path,
+};
 
 use thiserror::Error;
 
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @IamTheCarl, looks good!

> except for 3MF which needs the target to implement both Write and Seek

I wonder why that is, and if that's a real requirement. Maybe it's just a weird leftover from an earlier version. Don't know!

IamTheCarl:
I wondered that too so I gave the source code a quick peruse (I would have updated the ThreeMF library if it looked like an easy fix)

ThreeMF uses the Zip crate to [produce the 3MF file](https://github.com/hannobraun/3mf-rs/blob/main/src/write.rs#L17).
ZipWriter expects an object that is [both Write and Seek](https://docs.rs/zip/0.6.6/zip/write/struct.ZipWriter.html#method.new)
ZipWriter uses the seek functionality of the writer multiple times during the finalization process.

This isn't a big problem for me though. I can just output the 3MF to a tempfile before I send it to stdout, in the rare case that someone wants to pipe a 3MF file to stdout. The only situation I can see that being useful in is if someone wanted to automatically pipe a model into a slicer without an intermediate file to clean up later, which is actually something I can see myself doing.

hannobraun:
That makes sense. Thank you for looking into it!

> This isn't a big problem for me though. I can just output the 3MF to a tempfile before I send it to stdout, in the rare case that someone wants to pipe a 3MF file to stdout. The only situation I can see that being useful in is if someone wanted to automatically pipe a model into a slicer without an intermediate file to clean up later, which is actually something I can see myself doing.

You actually don't need to go through a file. You can use [`Cursor`](https://doc.rust-lang.org/std/io/struct.Cursor.html) to do everything in-memory.

IamTheCarl:
I did consider that but my concern is how much memory do I need? I'm probably being paranoid (thanks to my embedded background) but a spooled temp file is the best of both worlds and easy to use: https://docs.rs/tempfile/3.10.1/tempfile/struct.SpooledTempFile.html

hannobraun:
Ah, okay. I just mentioned it in case you're not aware. `SpooledTempFile` seems like a safe bet.

PR #2253, opened 2024-03-06
Update with arbitrary containers of Handles or HandleWrappers

IamTheCarl:
The main purpose of this PR was to make it easier to check for duplicate handles referencing the same object, however it comes with a useful side effect.

I personally needed this because my CAD program accepts a scripting language to generate models with. The number of changes a user may make is dynamic, so I need the option to provide a variable number of new faces when updating a shell. I also need to validate that the new list of faces does not contain any duplicate handles so that I can avoid a panic and produce a more meaningful error message for the user.

The key change is that `ObjectSet::new` and `ObjectSet::replace` no longer take just arrays of Handles, but instead they take any iterable object that produces objects that can be turned into HandleWrapper.

This provides two key features:
1. You can pass an array of either Handle or HandleWrapper. Handle Wrapper makes it much easier to check if your handles are duplicates.
2. Any iterable object, such as a vector can be used for the update. I can only know how many shells my user will produce at runtime, so this is useful to me.

With such a workflow, things like this are now easy to implement:
```rust
    let mut result = Ok(());

    solid.update_shell(
        solid.shells().only(),
        |shell, core| {
           let new_faces: Vec<HandleWrapper<Face>> =
              vec![ /* The user does whatever they need to do to generate more faces */ ];
           
           // A HashSet is probably not an ideal way to search for duplicates on this small of a scale,
           // but it's good enough for this demo.
           let new_face_set = HashSet::from(new_faces.iter().map(|face_handle| face_handle.id()));
           
           if new_face_set.len() == new_faces.len() {
               new_faces // Nothing is wrong, we are clear to continue.
           } else {
               // This is an error situation.
               result = Err(Failure::DuplicateFaces);
               
               // Because there currently isn't a way to abort an update, we'll just return a copy of our
               // original shell this results in an unmodified solid.
               [shell.clone()]
           }
        },
        core,
    );

    return result;
```

I think at this point it might be good to rename `HandleWrapper` to `ObjectReference` to imply that we are comparing the references and not the objects, but I'll let you make that call. Coming up with a name that reflects the purpose of that thing is... difficult.

### DIFF
--- a/crates/fj-core/src/objects/object_set.rs
+++ b/crates/fj-core/src/objects/object_set.rs
@@ -28,16 +28,17 @@ impl<T> ObjectSet<T> {
     /// # Panics
     ///
     /// Panics, if the iterator contains duplicate `Handle`s.
-    pub fn new(handles: impl IntoIterator<Item = Handle<T>>) -> Self
+    pub fn new<H>(handles: impl IntoIterator<Item = H>) -> Self
     where
         T: Debug + Ord,
+        H: Into<HandleWrapper<T>>,
     {
         let mut added = BTreeSet::new();
         let mut inner = Vec::new();
 
-        for handle in handles {
-            let handle = HandleWrapper::from(handle);
+        let handles = handles.into_iter().map(|handle| handle.into());
 
+        for handle in handles {
             if added.contains(&handle) {
                 panic!(
                     "Constructing `ObjectSet` with duplicate handle: {:?}",
@@ -149,13 +150,14 @@ impl<T> ObjectSet<T> {
     ///
     /// Panics, if the update results in a duplicate item.
     #[must_use]
-    pub fn replace<const N: usize>(
+    pub fn replace<H>(
         &self,
         original: &Handle<T>,
-        replacements: [Handle<T>; N],
+        replacements: impl IntoIterator<Item = H>,
     ) -> Option<Self>
     where
         T: Debug + Ord,
+        H: Into<Handle<T>>,
     {
         let mut iter = self.iter().cloned().peekable();
 
@@ -185,7 +187,7 @@ impl<T> ObjectSet<T> {
         Some(
             before
                 .into_iter()
-                .chain(replacements)
+                .chain(replacements.into_iter().map(|handle| handle.into()))
                 .chain(after)
                 .collect(),
         )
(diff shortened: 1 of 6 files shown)


crates/fj-core/src/objects/object_set.rs
```diff
@@ -28,16 +28,17 @@ impl<T> ObjectSet<T> {
     /// # Panics
     ///
     /// Panics, if the iterator contains duplicate `Handle`s.
-    pub fn new(handles: impl IntoIterator<Item = Handle<T>>) -> Self
+    pub fn new<H>(handles: impl IntoIterator<Item = H>) -> Self
     where
         T: Debug + Ord,
+        H: Into<HandleWrapper<T>>,
```
hannobraun, line 34: This can be written a bit more compactly:

```suggestion
    pub fn new(handles: impl IntoIterator<Item = impl Into<HandleWrapper<T>>) -> Self
    where
        T: Debug + Ord,
```

This is the style I usually prefer, but it's not actually that much better, in any meaningful way. I'll merge this anyway, but I figured I'd mention the possibility, in case you weren't aware.

crates/fj-core/src/objects/object_set.rs
```diff
@@ -149,13 +150,14 @@ impl<T> ObjectSet<T> {
...
-    pub fn replace<const N: usize>(
+    pub fn replace<H>(
         &self,
         original: &Handle<T>,
-        replacements: [Handle<T>; N],
+        replacements: impl IntoIterator<Item = H>,
     ) -> Option<Self>
     where
         T: Debug + Ord,
+        H: Into<Handle<T>>,
```
hannobraun, line 160: Same comment about `ObjectSet::new` applies here too.

crates/fj-core/src/objects/object_set.rs
```diff
@@ -260,4 +267,38 @@ mod tests {
...
+    fn deduplicate_with_hashset() {
+        let mut core = Core::new();
+
+        let bare_cycle = Cycle::new([]);
+        let cycle = bare_cycle.insert(&mut core);
+
+        let standard_set = HashSet::from([cycle.clone(), cycle.clone()]);
+
+        assert_eq!(standard_set, [cycle].into());
+    }
```
hannobraun, line 281: I don't think this belongs here. It doesn't actually test `ObjectSet`. I'd suggest either removing it, or adapting it as an example in the documentation of `ObjectSet::new`.

I'm going to merge the pull request anyway, as I don't think this is a big deal. Feel free to send a follow-up PR, if you want to. If not, I'll most likely remove it in a future cleanup pass.

hannobraun (approved):
Thank you, @IamTheCarl! I left a few comments to complain about this or that, but overall it looks great. Merging.

IamTheCarl:
I'll see about applying these comments tonight, but I'm glad you overall like it.

PR #2254, opened 2024-03-06
Made `export_obj` public

IamTheCarl:
I went to use the new export features in Command CAD and discovered I had made a very small mistake.


hannobraun (approved):
Thank you, @IamTheCarl!

PR #2255, opened 2024-03-06
Apply suggested cleanup from #2253

IamTheCarl:
I don't think additional explanation will be helpful here.


hannobraun (approved):
Thank you, @IamTheCarl! I appreciate you following up!

Issue #2158, opened 2024-01-15
Validate winding of sketches


hannobraun:
`Sketch`es each contain multiple `Region`s. The exterior cycle of those regions should have counter-clockwise winding, their interior cycles should have clockwise winding. I believe that the sweep code already assumes this, but currently, this is not documented nor enforced.

## Implementation

This new validation check would go into the `validate` module, specifically [`validate::sketch`](https://github.com/hannobraun/fornjot/blob/c93d4a1c5d987b0cbb1694ecf235be8cb8369c42/crates/fj-core/src/validate/sketch.rs). You can use the structure of the existing validation check (for `Sketch` and other objects) as a guideline for the implementation.

The check itself should be rather straight-forward, as it could just use [`Cycle::winding`](https://docs.rs/fj-core/0.48.0/fj_core/objects/struct.Cycle.html#method.winding). Please note that this method is buggy (https://github.com/hannobraun/fornjot/issues/2130), but as long as we make sure not to use any complicated cycles in the unit test for this validation check, we'll be fine (just stick to polygons).

Since this should be relatively straight-forward, and there are already numerous validation checks to serve as inspiration, I'm designating this as a https://github.com/hannobraun/fornjot/labels/good%20first%20issue.

## Related Issues

There are two related issues:

- https://github.com/hannobraun/fornjot/issues/2132, which would change the structure of regions (implementing that first would simplify this issue).
- https://github.com/hannobraun/fornjot/issues/2157, which would restructure the validation infrastructure (implementing that first would change where this check would go, and some details about how it would be implemented).

Neither of those issues are blockers for this one, but work would need to be coordinated. If you're interested in picking this up, I suggest you check out those issues to see if anybody is already working on them.

hannobraun:
Addressed by @emka in https://github.com/hannobraun/fornjot/pull/2256.

Issue #2130, opened 2023-12-11
`Cycle::winding` makes assumptions that don't hold for all cycles


hannobraun:
[`Cycle::winding`](https://github.com/hannobraun/fornjot/blob/3e3e9920df98db5d23ca956d2d0282baca35590f/crates/fj-core/src/objects/kinds/cycle.rs#L32-L81) is a method that computes the winding of a cycle (as the name suggests), meaning it determines whether a cycle's half-edges are oriented clockwise or counter-clockwise. The method works well for all cycles made up of straight half-edges (i.e. polygons) and circles. It does not work for all cycles that contain arcs, as it makes some assumptions that don't always hold in the presence of those.

Exhibit A:
https://github.com/hannobraun/fornjot/blob/3e3e9920df98db5d23ca956d2d0282baca35590f/crates/fj-core/src/objects/kinds/cycle.rs#L46-L51

Imagine a half-circle whose end-points are connected by a line segment. Totally valid, but unless the arc happens to be the first half-edge in the cycle, it will trigger this panic.

Exhibit B:
https://github.com/hannobraun/fornjot/blob/3e3e9920df98db5d23ca956d2d0282baca35590f/crates/fj-core/src/objects/kinds/cycle.rs#L61-L80

The comment is wrong! Imagine a crescent shape. This code treat that like two coincident line segments, meaning `sum` will be zero and we'll run into the panic.

---

There might be more issues, but those are two that I happened to discover. I believe this once worked as designed, when we were limited to line segments and full circles. Once arcs were introduced, the assumptions here no longer held.

I think it makes sense to hold off on fixing this. Nobody has complained about this so far, and if https://github.com/hannobraun/fornjot/issues/2118 works out, fixing it will become trivial.

PR #2256, opened 2024-03-10
Add winding validation for sketches

emka:
Tests use circles, arcs currently can not be tested (#2130). Adding sketch validation for #2158.

### DIFF
--- a/crates/fj-core/src/validate/sketch.rs
+++ b/crates/fj-core/src/validate/sketch.rs
@@ -54,7 +54,9 @@ mod tests {
     use crate::{
         assert_contains_err,
         objects::{Cycle, HalfEdge, Region, Sketch, Vertex},
-        operations::{build::BuildHalfEdge, insert::Insert},
+        operations::{
+            build::BuildHalfEdge, build::BuildRegion, insert::Insert,
+        },
         validate::{
             references::ReferenceCountError, SketchValidationError, Validate,
             ValidationError,
@@ -66,15 +68,15 @@ mod tests {
     fn should_find_cycle_multiple_references() -> anyhow::Result<()> {
         let mut core = Core::new();
 
-        let shared_cycle = Cycle::new(vec![]).insert(&mut core);
+        let region = <Region as BuildRegion>::circle([0., 0.], 1., &mut core)
+            .insert(&mut core);
+        let valid_sketch = Sketch::new(vec![region.clone()]).insert(&mut core);
+        valid_sketch.validate_and_return_first_error()?;
 
+        let shared_cycle = region.exterior();
         let invalid_sketch = Sketch::new(vec![
-            Region::new(
-                Cycle::new(vec![]).insert(&mut core),
-                vec![shared_cycle.clone()],
-            )
-            .insert(&mut core),
-            Region::new(shared_cycle, vec![]).insert(&mut core),
+            Region::new(shared_cycle.clone(), vec![]).insert(&mut core),
+            Region::new(shared_cycle.clone(), vec![]).insert(&mut core),
         ]);
         assert_contains_err!(
             invalid_sketch,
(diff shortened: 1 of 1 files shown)


emka:
Test cases `should_find_cycle_multiple_references` and `should_find_half_edge_multiple_references` are currently failing because they contain cycles without valid winding. Do you have a preference how to handle this case?

crates/fj-core/src/validate/sketch.rs
```diff
@@ -129,4 +172,79 @@ mod tests {
...
+            &valid_outer_circle,
+            Vertex::new().insert(&mut core),
+        )
+        .insert(&mut core);
+        let invalid_exterior =
+            Cycle::new(vec![invalid_outer_circle.clone()]).insert(&mut core);
+        let invalid_sketch =
+            Sketch::new(vec![
+                Region::new(invalid_exterior.clone(), vec![]).insert(&mut core)
+            ]);
```
hannobraun, line 200: This can probably be shortened by updating the valid sketch. Something like this:
``` rust
let invalid_sketch = valid_sketch.update_region(valid_sketch.regions().first(), |region|
    region.update_exterior(|cycle| cycle.reverse())
);
```

I didn't test this, so this might not work precisely like this. But something along those lines should be possible.

crates/fj-core/src/validate/sketch.rs
```diff
@@ -129,4 +172,79 @@ mod tests {
...
+        .insert(&mut core);
+        let exterior = Cycle::new(vec![outer_circle.clone()]).insert(&mut core);
+
+        let valid_interior =
+            Cycle::new(vec![cw_inner_circle.clone()]).insert(&mut core);
+        let valid_sketch = Sketch::new(vec![Region::new(
+            exterior.clone(),
+            vec![valid_interior],
+        )
+        .insert(&mut core)]);
```
hannobraun, line 232: Can probably be shortened by using available APIs. Maybe (attention, pseudo-code) `Region::circle().add_interior(invalid_circle)`.

crates/fj-core/src/validate/sketch.rs
```diff
@@ -129,4 +172,79 @@ mod tests {
...
+        valid_sketch.validate_and_return_first_error()?;
+
+        let invalid_interior =
+            Cycle::new(vec![inner_circle.clone()]).insert(&mut core);
+        let invalid_sketch = Sketch::new(vec![Region::new(
+            exterior.clone(),
+            vec![invalid_interior],
+        )
+        .insert(&mut core)]);
```
hannobraun, line 241: Can probably be shortened using `Sketch::update_region` and other operations, as suggested above in the other test.

hannobraun (changes requested):
Thank you, @emka, great work!

I left some comments with suggestions. But to be clear, none of those need to be addressed. Feel free to look into them, if you're up for it. But I'm happy to leave this as-is, as the code can always be cleaned up later, when making changes to it.

What needs to be addressed is the CI failure.


> Test cases `should_find_cycle_multiple_references` and `should_find_half_edge_multiple_references` are currently failing because they contain cycles without valid winding. Do you have a preference how to handle this case?

My preferred approach to these tests is to build a valid object using high-level builder APIs, then create an invalid variant by doing a minimal update of the valid object. Just like I've suggested for the tests you added.

I'm not too picky, however. Looks like the failing tests can be fixed by just replacing `Cycle::new` with `Cycle::circle`, so feel free to just do that!

emka:
Thank you for the feedback! I want to take care of the unresolved comments as well before merging.

hannobraun:
Great, thank you! Let me know when this is ready for another look.

One additional note: It would be great, if the commits with changes to the existing tests came *before* the commit with new validation check, to make sure every single commit builds/tests correctly (this can be very useful when using `git bisect`). (In case you're not familiar, `git rebase -i` is probably the best tool to achieve this. If you're having problems, let me know. I can also do this on my side, if necessary.)

hannobraun (approved):
Hey @emka, I've decided to merge this pull request right away. I'm about to make some changes to the validation code that are not directly related to this pull request, but might touch some of the same code. Keeping this open will only lead to merge conflicts and frustration.

I'll do a rebase before merging, to get the commits in the order I requested.

I'd be happy to accept further improvements in follow-up pull requests!

Issue #2116, opened 2023-11-29
Separate geometry from topology


hannobraun:
As far as I know, b-rep CAD kernels traditionally separate topology (how objects relate to each other) from geometry (where objects are). At least that's what I got from Boundary Representation Modelling Techniques by Ian Stroud. Earlier on, Fornjot followed that separation, but that didn't survive the many improvements and refactorings since then.

I think it's time to bring that idea back, but with a different twist: Originally the topological data referred to the geometry (and that's how it sill works, kinda, but without a clear separation). I'd like to do it the other way around: Geometrical data exists as a separate layer that refers to the topological object graph, which itself is isolated and doesn't concern itself with anything else. This is possible, because the (topological) objects in the graph have unique IDs, and thus can be referenced easily by an outside entity.

I see the following advantages in this arrangement:

- **Simplified object graph:** I'd expect the object graph to survive this transition mostly unchanged, but some objects will lose some of their attributes, simplifying the graph overall.
- **Easier to update geometry:** Topological objects are immutable and stored in a centralized data structure. Updating an object means creating a new version of it (as well as new versions of all objects that reference it), and storing those new versions.
This is not always appropriate. For example, if you have a user drag a vertex in a GUI, you'd have to create new versions of those objects every pixel.
- **Easier to evolve geometry:** The geometric primitives in Fornjot are currently very basic. Evolving that in the current structure would mean evolving the object graph with it, creating friction in the development process.
Having geometry be separate, would make it easier to experiment with geometry, as changes there would no longer affect the object graph. It would even be possible to have multiple competing geometry layers, and experiment with them in parallel.

Initially, the geometry layer could be a struct (`Geometry`), that is passed into all operations (and other code) that need it. Eventually, we should find a better way to organize this, as to not have a million different arguments in every function. I already have some ideas here, but that's a topic for another day (I have a long list of issues I need to open, so it might take me a bit to get to that one).

Unless I'm missing something (which is quite possible), this should be a relatively straight-forward refactoring. We should definitely attempt to do this before making any other significant changes to the geometry representation or object graph.

hannobraun:
#2117, which I wanted to do as a test run for this issue, is completed now. I will now start working on this one.

hannobraun:
As I'm starting to get into this, I've decided that I'm going to migrate surface geometry to the new geometry layer first. This has already turned up two issues that need handling:

[...]

PR #2267, opened 2024-03-18
Remove redundant validation check

hannobraun:
This check essentially verified that the length of a half-edge is non-zero, but I don't think this is actually important. At least I can't think of a reason why it wouldn't work, and there might be legitimate use cases.

What is important, is whether the vertices that bound the half-edge are coincident, but in this regard, this check could have lead to false positives, as the half-edge could be bounded by the same vertex on both sides.

In addition, there are other validation checks that check for coincident vertices, either existing in code, or tracked in issues.

### DIFF
--- a/crates/fj-core/src/validate/edge.rs
+++ b/crates/fj-core/src/validate/edge.rs
@@ -1,5 +1,3 @@
-use fj_math::{Point, Scalar};
-
 use crate::{
     geometry::Geometry,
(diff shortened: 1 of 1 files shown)


hannobraun:
Removing this validation check also un-complicates what I'm currently working on for https://github.com/hannobraun/fornjot/issues/2116, which is how I got to thinking about this validation check in the first place.

PR #2409, opened 2024-07-07
Update cube.obj

thestig4242:
applied material textures in orientations so the top view is default.


hannobraun (approved):
Thank you, @thestig4242! Works perfectly!

Issue #2424, opened 2024-07-22
Upgrade to wgpu 22


hannobraun:
This upgrade requires manual intervention. Here's the build failure from the automatic upgrade attempt in https://github.com/hannobraun/fornjot/pull/2422:
https://github.com/hannobraun/fornjot/actions/runs/10040783745/job/27747509273?pr=2422

Labeling https://github.com/hannobraun/fornjot/labels/good%20first%20issue, as this doesn't require a deep knowledge of Fornjot, those build errors don't look too intimidating, and the recent wgpu upgrades haven't been too involved.

felix91gr:
Lemme try it. If I don't ping back in 24 hours, feel free to assign it to someone else c:

felix91gr:
Okay, I have a candidate fix. Let me test it and then make a PR

felix91gr:
I might need some advice. Here is what I have for now.

### Summary

I have fixes for the 3 build errors that appeared in CI. The three of them, I think, can be dealt with without much fuss.

Further testing does reveal another build error, which I'll describe in the second section.

### Fixes for the 3 errors seen in CI

These are the errors, reproduced on my end:

![image](https://github.com/user-attachments/assets/1749e01f-3e12-4602-8912-65fec950d71f)

The first one is due to the new `MemoryHints` field that is fed to `DeviceDescriptor`. I am currently asking for a `MemoryUsage` profile, on the grounds that it should help with compatibility (I think, I might be wrong on this).

This error is fixed by just adding a memory usage profile.

The second and third errors are due to a new field, `cache: Option<PipelineCache>` that is now a part of `RenderPipelineDescriptor`. I have set it to `None` for now, and can be changed later if more performance is needed.

Those fixes make `cargo build`, `test` and `clippy` all pass without a problem. 

### The new issue I'm looking at (advice is much appreciated)

I ran `just ci` locally, and this is the error I get:

![image](https://github.com/user-attachments/assets/80621200-5bd8-44cc-bd93-0157e44af059)

The `RUSTFLAGS` that were used are the following:

`Running RUSTFLAGS="-D warnings" "cargo" "build" "--all-features" "--target" "wasm32-unknown-unknown" "-p" "fj-viewer"`

I'm not sure what to do about this error.

Reading the release notes, it looks like `enumerate_adapters` is now native-only:

![image](https://github.com/user-attachments/assets/87be7789-0cb5-48af-b1d1-8556c4bb427f)

[...]

PR #2427, opened 2024-07-26
Upgrade to wgpu 22

felix91gr:
Solves #2424 

CC @hannobraun, how does this look? Let me know what you think :slightly_smiling_face:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -216,11 +216,11 @@ checksum = "175571dd1d178ced59193a6fc02dde1b972eb0bc56c892cde9beeceac5bf0f6b"
 
 [[package]]
 name = "ash"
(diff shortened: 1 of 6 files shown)


hannobraun (approved):
Thank you, @felix91gr, looks good!

I wonder if this could be a bit simpler, going with one of the following options instead of modifying `try_from_all_adapters`:

- Put the `#[cfg(...)]` on `try_from_all_adapters` as a whole (leaving its body unchanged) and at the call site.
- Have two versions of `try_from_all_adapters`, instead of the one with a relatively complex body.

But that's just me wondering. I'm not sure if either of those alternatives would be practical, or actually better.

Feel free to try, if you want, or not. I'm going to merge this now in any case.

Thank you!

Issue #2161, opened 2024-01-16
Improve shading to properly display blind holes


hannobraun:
## Problem

This is how blind holes currently look:
![Screenshot from 2024-01-16 11-28-41](https://github.com/hannobraun/fornjot/assets/85732/f390c30f-dbc0-4899-accb-b887fd31f5ca)

You can reproduce this picture, by running `cargo run -p holes` from the repository root, then rotating (drag while left mouse button is pressed) and moving (drag while right mouse button is pressed) the model until you see the blind hole.

As you can see in this screenshot, the surface of the part and the bottom of the blind hole look exactly the same. This is because they are parallel, and the shading only depends on the angle between a surface and the viewing direction.

## Solution

I'm not an expert on graphics, and I'm sure there are many ways to improve this. If you know of one, feel free to implement it!

The solution that comes to my mind, is to make shading also depend on the distance to the camera. This would have to be carefully tuned, to make cases as the one shown in the screenshot look decent, without making far-away parts completely black. It might make sense to clamp the result in such a way, that the effect of distance can never make a part too dark.

## Implementation

Rendering is implemented in the [`fj-viewer` crate](https://github.com/hannobraun/fornjot/tree/main/crates/fj-viewer), specifically within [`shader.wgsl`](https://github.com/hannobraun/fornjot/blob/c93d4a1c5d987b0cbb1694ecf235be8cb8369c42/crates/fj-viewer/src/graphics/shader.wgsl#L38-L51). It's possible that this issue can be completely addressed with only changes to the linked shader function. It's also possible (especially, if a more advanced solution is chosen), that this requires changes to the larger graphics system.

In any case, implementing this will only require knowledge shader languages (I'm guessing that WGSL is pretty easy to understand for anyone familiar with other shader languages), and possibly Rust and [wgpu](https://crates.io/crates/wgpu). It won't require knowledge of other Fornjot internals. For this reason, I'm tagging this as a https://github.com/hannobraun/fornjot/labels/good%20first%20issue.

felix91gr:
It's been a long while since I've done shader programming, but as far as I'm aware the canonical way that these kinds of shadows are computed and drawn is Shadow Mapping.

https://github.com/gfx-rs/wgpu-rs/tree/master/examples/shadow I'm going to see if I can extract the design from the wgpu-rs examples, and will ping you back if I do :)

felix91gr:
Okay I think I'm right regarding the shadow map, but currently I'm not able to work on this.

[...]

PR #2470, opened 2024-09-28
added the suggested solution, but not working currently

aabdullah27182845:
Currently trying to resolve issue #2161 (Improve shading to properly display blind holes). The current resolution is not working as intended, but aiming to resolve this issue. This is why I have kept this as a draft pull request. Thanks!

### DIFF
--- a/crates/fj-viewer/src/graphics/shader.wgsl
+++ b/crates/fj-viewer/src/graphics/shader.wgsl
@@ -40,7 +40,7 @@ fn frag_model(in: VertexOutput) -> FragmentOutput {
     let light = vec3<f32>(0.0, 0.0, -1.0);
     let distance_from_camera = length(in.position.xyz);
 
-    let darkening_factor = 0.5 + 1.0/ (1.0 + exp(distance_from_camera));
+    let darkening_factor = 560.0 / (560.0 + exp(0.006 * distance_from_camera));
 
     let angle = acos(dot(light, -in.normal));
     let f_angle = angle / (pi * 0.75);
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you, @aabdullah27182845! As I said in the issue, I'll be able to take a closer look next week.

aabdullah27182845:
Thanks @hannobraun !

aabdullah27182845:
For anyone else who's interested in looking deeper into this, I can list the following findings. 

Adding a shader that depends on the distance from the initialised light point isn't the way to go about this, since we get interference on the shading of the surface of the cuboid of the surface, not just the hole.

With my limited knowledge on wgsl and limited knowledge on vector algebra, I think the way this could be solved is by calculating the normal vector to the plane of the cuboid, and then using that vector as our base and multiplying a darkening factor on top of that. Normal vectors always tend to be perpendicular to the plane, so I think that would be something reasonable we would use here.

Also, I'm in no way a computer graphics expert at all, I just thought this was a cool problem and wanted to solve it.
Screenshot from 2024-09-28 23-42-51


The picture above is what the normal vector is to the cuboid (perpendicular to the cuboid). I'm suggesting we solve that (I don't know how in the code, but usually they are solved using dot products equalling zero), and then use that vector to scale the shading.

hannobraun:
The code already does what you're suggesting. It is computing a shading factor based on the angle to the normal here: https://github.com/hannobraun/fornjot/blob/ef5830fc79db39d64c5747968ff0eab37895b940/crates/fj-viewer/src/graphics/shader.wgsl#L45

This is visible in your screenshot. See the side walls of the holes.

That alone is not enough though, due to the problem tracked in https://github.com/hannobraun/fornjot/issues/2161.

hannobraun:
Closing due to inactivity.

@aabdullah27182845, feel free to reopen this pull request (or maybe open a new one) if you want to pick this up again in the future.

Issue #2507, opened 2025-03-16
Feature Request: Support for hex color codes in Color


MilliUser:
### Problem
Hexadecimal color codes (`#RRGGBB` or `#RRGGBBAA`) are widely used in web development and design tools. Currently, the `Color` struct does not support creating colors directly from hex strings, which limits its usability in some scenarios.

### Proposed Solution
I propose adding a `From<&str>` implementation to the `Color` struct to allow conversion from hex color codes. This would enable users to easily create colors from hex strings like `#FF5733` or `#FF573380`.

### Example Implementation
Here is how the implementation could look:
crates/fj-interop/src/color.rs
```rust
...
impl From<&str> for Color {
    fn from(hex: &str) -> Self {
        let hex = hex.trim_start_matches('#');
        let len = hex.len();

        let parse_hex = |start: usize, end: usize| -> u8 {
            u8::from_str_radix(&hex[start..end], 16).unwrap_or(0)
        };

        match len {
            6 => {
                let r = parse_hex(0, 2);
                let g = parse_hex(2, 4);
                let b = parse_hex(4, 6);
                Self([r, g, b, 255])
            }
            8 => {
                let r = parse_hex(0, 2);
                let g = parse_hex(2, 4);
                let b = parse_hex(4, 6);
                let a = parse_hex(6, 8);
                Self([r, g, b, a])
            }
            _ => Self::default(),
        }
    }
}
```
### Benefits
This feature would simplify color creation from hex codes, which are commonly used in web and graphic design. It would also improve interoperability with other tools and libraries that use hex colors.

hannobraun:
Thank you for opening this issue, @MilliUser. This request makes sense. I am not going to work on this though, as I'm still focused on the ongoing [series of experiments](https://github.com/hannobraun/fornjot/tree/main/experiments).

In principle, I'd be happy to accept a pull request in the meantime, but that may be wasted effort: At this point, it's unclear how much of the old code will survive the eventual merge of the experimental code back into the mainline. And in fact, I've been considering going the opposite direction with this: Declare color (and anything more advanced like textures) as out of scope for this library. This is something that can be built on top.

MilliUser:
Thank you for your response and for considering my request! I appreciate the transparency about the current focus on experiments and the potential future direction of the library. I completely understand that the codebase is undergoing significant changes, and I respect the decision to potentially keep features like color and textures out of scope for the core library.

[...]

PR #2526, opened 2025-05-20
feature-  Support for hex color codes in Color #2507

rajgandhi1:
added support for hex colors and tests to check for issues.

### DIFF
--- a/crates/fj-interop/src/color.rs
+++ b/crates/fj-interop/src/color.rs
@@ -38,3 +38,64 @@ impl From<[f64; 3]> for Color {
         Self::from([r, g, b, 1.])
     }
 }
+
+impl From<&str> for Color {
+    fn from(hex_str: &str) -> Self {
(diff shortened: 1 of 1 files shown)


hannobraun (changes requested):
Thank you for opening this pull request, @rajgandhi1!

This looks mostly good, but I don't like how it falls back to the default color, if it can't parse the string successfully. I just know this would be _very_ confusing to somebody at some point, wasting a lot of time for them :smile: 

I think it would be more appropriate to implement [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) instead.

rajgandhi1:
@hannobraun ya that sounds better. Will change it

rajgandhi1:
@hannobraun  check now

hannobraun (approved):
Looks great! Thank you!

Issue #2307, opened 2024-03-28
Request: Add variant of `process_model` that accepts `Args` rather than parsing them


starthal:
Fornjot already provides a `process_model` helper that accepts a "standard" set of arguments for exporting or viewing.

I want to make an app that uses arguments to configure/parameterize a model. But `process_model` calls `parse` directly so I can't extend the arguments with my own.

With an API like `fn process_model_with_args<M>(fj: &mut fj::Instance, model: &M, args: fj::Args) -> fj::Result` this would be straightforward.

For example:

```rust
#[derive(clap::Parser)]
pub struct Parameters {
    #[clap(required = true, long, value_parser = parse_vector_3)]
    /// Size of the cuboid, as a comma-separated vector `x,y,z`
    size: [f64; 3],
    #[command(flatten)]
    /// Standard fj args
    fj: fj::Args,
}

fn parse_vector_3(arg: &str) -> anyhow::Result<[f64; 3]> {
    Ok(arg
        .split(',')
        .map(str::parse)
        .collect::<Result<Vec<f64>, _>>()?
        .as_slice()
        .try_into()?)
}

fn main() -> anyhow::Result<()> {
    let mut fj = fj::Instance::new();
    let p = <Parameters as clap::Parser>::parse();
    let model = cuboid::model(p.size, &mut fj.core);
    process_model_with_args(&mut fj, &model, p.fj)?;
    Ok(())
}
```

The user can invoke this like:

```
cargo run -- --size 3,5,7 --export cuboid.stl
```

hannobraun:
Thank you for opening this issue, @starthal! That's a very reasonable request.

Happy to accept a pull request implementing what you propose. Otherwise, I'll get this done whenever I can get to it.

hannobraun:
Labeling as `good first issue`, as this is a pretty small change, and the solution is already laid out in the issue description.

hannobraun:
Addressed in https://github.com/hannobraun/fornjot/pull/2527.

PR #2527, opened 2025-05-21
feature- Add variant of process_model that accepts Args #2307

rajgandhi1:


rajgandhi1:
@hannobraun please review this

models/cuboid/src/main.rs
```diff
@@ -1,6 +1,31 @@
...
+}
+
+fn parse_vector_3(arg: &str) -> anyhow::Result<[f64; 3]> {
+    Ok(arg
+        .split(',')
+        .map(str::parse)
+        .collect::<Result<Vec<f64>, _>>()?
+        .as_slice()
+        .try_into()?)
+}
```
hannobraun, line 21: Maybe it makes more sense to have separate arguments for each dimension, like `-x`, `-y`, `-z`? This would be more flexible for the user, who wouldn't have to override everything or nothing; and it would be simpler, not requiring the custom parse function.

models/cuboid/src/main.rs
```diff
@@ -1,6 +1,31 @@
...
-    let mut fj = fj::Instance::new();
-    let model = cuboid::model([3., 2., 1.], &mut fj.core);
-    fj.process_model(&model)?;
+use clap::Parser;
+use fj::{Args, Instance};
+
+#[derive(Parser)]
+struct Parameters {
+    /// Size of the cuboid, as a comma-separated vector `x,y,z`
+    #[arg(long, value_parser = parse_vector_3, default_value = "1,1,1")]
```
hannobraun, line 7: Please change the new defaults to match the old ones. Having different sizes for the different dimensions makes sense for a test model, as it can make it easier to recognize where a mistake is, or spot the mistake in the first place.

models/holes/src/lib.rs
```diff
@@ -13,7 +13,7 @@ pub fn model(radius: impl Into<Scalar>, core: &mut fj::core::Core) -> Solid {
     let radius = radius.into();
 
     let size = radius * 4.;
-    let cuboid = cuboid::model([size * 2., size, size], core);
+    let cuboid = ::cuboid::model([size * 2., size, size], core);
```
hannobraun, line 16: Why this change? It doesn't seem to make a difference, except to add visual noise.

models/cuboid/Cargo.toml
```diff
@@ -3,5 +3,11 @@ name = "cuboid"
 version = "0.1.0"
 edition = "2021"
 
-[dependencies.fj]
-path = "../../crates/fj"
+[[bin]]
+name = "cuboid"
+path = "src/main.rs"
```
hannobraun, line 8: Why add this? This is the default behavior anyway, and I don't see a benefit in having it.

models/cuboid/Cargo.toml
```diff
@@ -3,5 +3,11 @@ name = "cuboid"
...
+
+[dependencies]
+fj = { path = "../../crates/fj" }
```
hannobraun, line 11: Why change the formatting for the existing dependency, in addition to adding the new ones? Personally, I like the previous style better, as it's a bit less dense, which makes it a bit easier to read for me.

But it's a matter of taste. I wouldn't have complained if just the new dependency were formatted that way, but adding new code and making changes to existing in the same commit, makes this harder to review than it needs to be. If you want to make changes to existing code that aren't directly related, I would prefer having that in a separate commit, with an explanation on why the change is made.

models/holes/Cargo.toml
```diff
@@ -3,8 +3,12 @@ name = "holes"
 version = "0.1.0"
 edition = "2021"
 
-[dependencies.fj]
-path = "../../crates/fj"
+[[bin]]
+name = "holes"
+path = "src/main.rs"
```
hannobraun, line 8: Same comment from `cuboid/Cargo.toml` applies here to.

models/holes/Cargo.toml
```diff
@@ -3,8 +3,12 @@ name = "holes"
...
 
-[dependencies.cuboid]
-path = "../cuboid"
+[dependencies]
+fj = { path = "../../crates/fj" }
```
hannobraun, line 11: Same comment from `cuboid/Cargo.toml` applies here too.

models/holes/Cargo.toml
```diff
@@ -3,8 +3,12 @@ name = "holes"
...
+clap = { version = "4.5", features = ["derive"] }
+anyhow = "1.0"
```
hannobraun, line 13: I don't think the `anyhow` dependency is actually used in this model.

models/cuboid/src/main.rs
```diff
@@ -1,6 +1,31 @@
-fn main() -> fj::Result {
-    let mut fj = fj::Instance::new();
-    let model = cuboid::model([3., 2., 1.], &mut fj.core);
-    fj.process_model(&model)?;
+use clap::Parser;
+use fj::{Args, Instance};
```
hannobraun, line 2: The same comment from above, about making unrelated changes making things harder to review, applies here too.

In addition, I think this change makes the code clearly less readable here. The name of the `fj` crate is purposely short, to allow using things directly with the prefix. Now, it's harder to see where these types comes from, which is especially relevant because `Args` and `Instance` by themselves are such generic names.

hannobraun (changes requested):
Thank you for opening this pull request, @rajgandhi1!

I don't like how the new method duplicates most of the code from the existing `process_model` method. This is likely going to result in mistakes, if changes become necessary, and one of the copies gets missed.

Shouldn't it be possible to call the new method from `process_model`? It does all the same things, except parse the arguments, right?

---

I found some additional issues and left comments there. I stopped reviewing after the first 1-2 models, because the problems seemed to be the same in all of them.

models/cuboid/src/main.rs
rajgandhi1: got it. That way they can have default values for others.

hannobraun (approved):
Thank you again, for making more changes. Looks good now!

PR #2565, opened 2025-11-15
Generate Docs

bigmark222:
Generate cargo docs for crates overview.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,32 @@
+name: Build and Deploy Docs
+
+on:
+  push:
+    branches: [main]
+
+permissions:
+  contents: write
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Build developer documentation
+        run: cargo doc --workspace --all-features --no-deps
+
+      - name: Remove leftover lock file
+        run: rm -f target/doc/.lock
```
hannobraun: Why is this necessary? It seems like a workaround for some specific problem, and that problem should be documented here as a comment.

.gitignore
```diff
@@ -1,6 +1,14 @@
 # Cargo
 /target
 
+# Generated documentation
+developer_docs/
+docs/book/
+docs/build/
+docs/dist/
+public/
+site/
```
hannobraun: Why is this necessary? I thought documention is generated within `/target`, which is already ignored.

